PR #45, opened 2025-06-18
doc: Add shape optimization demo

dionhaefner:
#### Relevant issue or PR
n/a

#### Description of changes
This pulls in the shape optimization demo from our SciPy proceedings paper and enables testing on CI + rendering in the docs.

#### Testing done
manual + CI


examples/fem-shapeopt/fem_tess/tesseract_environment.yaml
```diff
@@ -0,0 +1,22 @@
+name: jax-fem-env
+channels:
+  - conda-forge
+  - defaults
```
jacanchaplais, line 4: recommend removing this, since you don't use any packages from Anaconda's commercial channel

examples/fem-shapeopt/demo.ipynb
jacanchaplais, line 1: L23 of the first cell should have "signed distance field (SDF)", but is missing the initialism in brackets. This initialism is actually introduced with the brackets later under "Define mapping from SDF to Density Field".

jacanchaplais, line 1: Some comments in the `visualize()` function would help me understand what's going on a bit better. For instance, I see you are using `sum(geometries, start=pv.PolyData())`. So I guess `geometries` is some iterable of `pv.PolyData` instances, but it's unclear to me what happens when you add them together. Is the result a reduction over array-like objects, or does it concatenate them, or something else?

.github/workflows/test_examples.yml
```diff
@@ -5,6 +5,10 @@ on:
     branches:
       - main
 
+  pull_request:
+    paths:
+      - examples/**
+
```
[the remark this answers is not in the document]
dionhaefner: Tesseract-JAX only tests tutorials on pushes to `main` instead of on every commit to keep feedback times reasonable on CI.

PR #47, opened 2025-06-20
feat: support simple sequential batching rule (support for `jax.vmap` and `jax.jacobian`)

### LINKED ISSUE #1: Add batching rule to support `vmap`

jpbrodrick89:
#### Relevant issue or PR
Fixes #1

#### Description of changes
Implement a simple sequential batching rule enabling the use of `vmap` and `jax.jacobian` by simply calling the relevant endpoint multiple times. The implementation was directly lifted from [jax._src.ffi.ffi_batching_rule](https://github.com/jax-ml/jax/blob/main/jax/_src/ffi.py#L638) with refactoring of variable names and eliminating dead code. No private JAX API is used; instead `split_args` and `unflatten_args` were refactored to serve additional purposes. The only thing I'm really not sure about is removing the update of the `output_layout` which I have just commented out (the relevant code now refers to  `output_pytreedef_expanded`).

Should we provide any attributions to `jax` in accordance with their Apache 2.0 [license](https://github.com/jax-ml/jax/blob/main/LICENSE) somewhere?

#### Testing done
- [x] CI passes
- [x] `jax.jacfwd` and `jax.jacrev` runs and produces expected results with `vectoradd_jax`
- [x] Added CI tests of `jacobian` 
- [x] Manually test more exotic uses of `vmap` and then add selection to CI

### Further work
In the future we can look at adding alternative `vmap_method`s: the main ones I think would be relevant are `async` (like sequential but we allow workers to run in parallel when this support is eventually added) and `auto` (expand dimensions where `InputSchema` has an ellipsis in the shape definition). For tangent/cotangent vectors I think we should actually consider supporting (or better yet requiring) our `jvp`/`vjp` endpoints to accept a batch of vectors as default. `jacobian` calculations using the sequential method will be slow for large array sizes and there is no way to tell `jax` to use our `jacobian` endpoint instead.


tesseract_jax/primitive.py
```diff
@@ -35,21 +37,13 @@ def __hash__(self) -> int:
...
-    array_args = tuple(
-        arg
-        for arg, is_static in zip(flat_args, is_static_mask, strict=True)  # fmt: skip
-        if not is_static
-    )
-
-    return array_args, static_args
+    flat_args: Sequence[T], mask: Sequence[bool]
+) -> tuple[tuple[T, ...], tuple[T, ...]]:
+    """Split a flat argument tuple according to mask (mask_False, mask_True)."""
```
jpbrodrick89, line 42: Happy to improve doc string, but I thought using masked/unmaked was confusing as I'm not even sure whether "masked" should refer to when mask values are zero or one!

tesseract_jax/primitive.py
```diff
@@ -35,21 +37,13 @@ def __hash__(self) -> int:
 
 
 def split_args(
```
jpbrodrick89, line 39: Not sure if we should move `split_args` (here) and `unflatten_args` (in `tesseract_compat`) into their own file (I won't suggest committing the cardinal sin of calling the file `utils` again! But I'd be happy to hear further suggestions! 😄

tesseract_jax/primitive.py
```diff
@@ -238,6 +232,55 @@ def _dispatch(*args: ArrayLike) -> Any:
...
+            output_avals=output_avals,
+            is_static_mask=is_static_mask,
+            client=client,
+            eval_func=eval_func,
+        )
+
+    g = lambda _, x: ((), _batch_fun(x))
+    _, outvals = jax.lax.scan(g, (), batched_args)
+
+    return tuple(outvals), (0,) * len(outvals)
```
jpbrodrick89, line 278: If anyone ever manages to find good (or any really) documentation for batching rules I'd love to work out what the second argument here is for without trawling through source code!

tesseract_jax/primitive.py
```diff
@@ -316,8 +359,10 @@ def apply_tesseract(
     client = Jaxeract(tesseract_client)
 
     flat_args, input_pytreedef = jax.tree.flatten(inputs)
+
```
jpbrodrick89, line 362:
```suggestion
```

jpbrodrick89:
@johnbcoughlin you might want to try this branch out along with the main version of `lineax`.

tesseract_jax/tesseract_compat.py
```diff
@@ -20,23 +29,14 @@ def unflatten_args(
...
-                combined_args.append(elem)
-
-        else:
-            combined_args.append(next(array_iter))
+    if remove_static_args:
+        static_args_converted = [None] * len(static_args)
+    else:
+        static_args_converted = [
+            elem.wrapped if hasattr(elem, "wrapped") else elem for elem in static_args
+        ]
```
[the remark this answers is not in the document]
dionhaefner: Let's reserve performance concerns for the things that matter :)

tesseract_jax/primitive.py
```diff
@@ -238,6 +232,55 @@ def _dispatch(*args: ArrayLike) -> Any:
...
+    new_args = [
+        arg if ax is batching.not_mapped else batching.moveaxis(arg, ax, 0)
+        for arg, ax in zip(array_args, axes, strict=True)
+    ]
+
+    # if output_pytreedef is not None:
+    #     output_pytreedef_expanded = tuple(
+    #         None if layout is None else tuple(n + 1 for n in layout) + (0,)
+    #         for layout in output_pytreedef
+    #     )
```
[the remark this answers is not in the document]
dionhaefner: This would be easier to discuss if we had some test coverage. Want to address that next, before we look at the code and potential failure modes in detail?

tesseract_jax/primitive.py
```diff
@@ -238,6 +232,55 @@ def _dispatch(*args: ArrayLike) -> Any:
...
+            input_pytreedef=input_pytreedef,
+            output_pytreedef=output_pytreedef,
+            output_avals=output_avals,
+            is_static_mask=is_static_mask,
+            client=client,
+            eval_func=eval_func,
+        )
+
+    g = lambda _, x: ((), _batch_fun(x))
+    _, outvals = jax.lax.scan(g, (), batched_args)
```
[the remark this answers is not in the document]
jpbrodrick89: This is how it's implemented in `ffi_batching_rule`, which means the docs are out of date. I'm guessing `scan` is now maybe smart enough to recognise this or they decided it didn't make a difference. I could dig into the scan implementation if you like and compare to map but I think we should just stress rest and see what happens.

tesseract_jax/primitive.py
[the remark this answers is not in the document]
jpbrodrick89: I missed that these lines are only evaluated in `ffi_batching_rule` for non-`sequential` batching rule, so should be safe if not necessary to remove. Therefore, I will remove these commented lines too.

tesseract_jax/primitive.py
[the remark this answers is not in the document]
jpbrodrick89: I found the [commit](https://github.com/jax-ml/jax/commit/c502332ed5976ec3c1aa4db6a2d06be7ecfb03fe) where this change was made and it was to support loop unrolling. I will revert to the `map` version for now, and we can bring `scan` back if we think unrolled may be better in the future. While unrolling can often be advantageous on GPU, ffi calls always take place on CPU so I don't think there'd be much if any advantage from unrolling.

jpbrodrick89:
@dionhaefner For now this should be enough to demonstrate the `vmap` rule works as expected for a variety of use cases with a single output. I will add `nested_tesseract` vmap tests in on Monday to see whether we get the right error if we abuse `out_axes`.

tests/test_endtoend.py
```diff
@@ -148,6 +150,90 @@ def f(x, y):
...
+@pytest.mark.parametrize("use_jit", [True, False])
+def test_univariate_tesseract_vmap(served_univariate_tesseract_raw, use_jit):
+    rosenbrock_tess = Tesseract(served_univariate_tesseract_raw)
+
+    # make things callable without keyword args
+    def f(x, y):
+        return apply_tesseract(rosenbrock_tess, inputs=dict(x=x, y=y))["result"]
+
+    # add one batch dimension
+    for axes in [(0, 0), (0, None), (None, 0)]:
```
dionhaefner, line 198: Can we test with a non-leading axis?

tesseract_jax/tesseract_compat.py
```diff
@@ -12,6 +13,14 @@
 PyTree: TypeAlias = Any
 
 
+def combine_args(args0: Sequence, args1: Sequence, mask: Sequence[bool]) -> tuple:
+    """Merge the elements of two lists based on a mask."""
```
dionhaefner, line 17: This could profit from an example in the docstring

tesseract_jax/primitive.py
dionhaefner: Could certainly be done, although it's not obvious to me what that module should be called. OK to me to leave as is.

tests/test_endtoend.py
```diff
@@ -148,6 +150,90 @@ def f(x, y):
     _assert_pytree_isequal(vjp, vjp_raw)
 
 
+@pytest.mark.parametrize("use_jit", [True, False])
+@pytest.mark.parametrize(
+    "jacfun", [partial(jax.jacfwd, argnums=(0, 1)), partial(jax.jacrev, argnums=(0, 1))]
```
dionhaefner, line 155: I suggest we use strings to distinguish between the cases, for better rendering when running the tests and to decouple the logic here a bit.

jpbrodrick89:
Thanks @dionhaefner I added a test over `out_axes` earlier this morning but didn't push, all worked fine there. I've been working on tests with nested Tesseract and getting a bit confused if I've implemented the right behaviour now. The current implementation will batch over every output even if not affected by the inputs. Is that what you'd expect as a user of vmap? Or should there be some way to change the unaffected outputs back to their original dimension with the standard `vmap` API (`None` doesn't work as the outputs have already been batched).

jpbrodrick89:
The other consideration I had was whether we should add any assertions that batch dims are equal or just rely on standard vmaps checks.

dionhaefner:
> Thanks @dionhaefner I added a test over `out_axes` earlier this morning but didn't push, all worked fine there. I've been working on tests with nested Tesseract and getting a bit confused if I've implemented the right behaviour now. The current implementation will batch over every output even if not affected by the inputs. Is that what you'd expect as a user of vmap? Or should there be some way to change the unaffected outputs back to their original dimension with the standard `vmap` API (`None` doesn't work as the outputs have already been batched).

There's no way to know which outputs are affected by which inputs, so yes this is the intended behavior.

jpbrodrick89:
@dionhaefner this should be good for the final sweep now. Agree that fixing the non-batched arguments is not possible without tracing through the Tesseract, maybe in the future we could add a `"differentiable_only"` option and then it only batches over `Differentiable` outputs which is probably the most common use case.

By the way the second argument of the batching rule is the axis that has been batched over. When vmapping this is referred to if needing to transpose to another axis.

With regards to supporting multiple tangent vectors as default which one of the three options below should I go with?:
1. Raise an issue to discuss options and point to on slack, perhaps leading to a mini-psync if hard to resolve
2. Open discussion on `dev-tesseract`
3. Outline ideas and justification in Confluence page and schedule a mini-psync.

tesseract_jax/tesseract_compat.py
```diff
@@ -12,6 +13,22 @@
...
 
+def combine_args(args0: Sequence, args1: Sequence, mask: Sequence[bool]) -> tuple:
+    """Merge the elements of two lists based on a mask.
+
+    The length of the two lists is required to be equal to the length of the mask.
+    `combine_args` will populate the new list according to the mask: if the mask evaluates
+    to `False` it will take the next item of the first list, if it evaluate to `True` it will
+    take from the second list.
+    For example, merging the lists ["foo", "bar"] and [0, 1, 2] wih the mask [1, 0, 0, 1, 1]
+    will return [0, "foo", "bar", 1, 2]
```
dionhaefner, line 24: We typically format examples like this:

```suggestion

    Example:
        >>> combine_args(["foo", "bar"], [0, 1, 2], [1, 0, 0, 1, 1])
        [0, "foo", "bar", 1, 2]
```

PR #71, opened 2025-08-21
fix: update version in vectoradd_jax example to not result in invalid docker tag

zmheiko:
#### Description of changes
Updated version in `vectoradd_jax` example to not result in invalid docker tag


zmheiko:
Yes, the demo notebook still runs through. I could not run the actual tests locally because of a weird docker issue I'm having (unrelated to the actual issue here), so I had to fall back to the GitHub runner to do the testing for me :)

PR #99, opened 2025-11-04
fix: partial derivatives + JIT causes error

### LINKED ISSUE #137: Traced arrays are always treated as differentiable
### Description

When using JIT compilation in conjunction with tesseract where certain inputs are non differentiable, this will cause an error because the vjp_inputs, jvp_inputs include an argument that is not supported. The main problem is:

```
def _is_static(x: Any) -> bool:
    if isinstance(x, jax.core.Tracer):
        return False
    return True
```

which also returns true if an array is traced because of JIT and not autodiff.

### Steps to reproduce

```
vectoradd_tess = Tesseract(served_vectoradd_tesseract)
    a = np.array([1.0, 2.0, 3.0], dtype="float32")
    b = np.array([4.0, 5.0, 6.0], dtype="float32")

    def loss_fn(a, b):
        vectoradd_fn_a: jax.Callable = lambda a: apply_tesseract(
            vectoradd_tess,
            inputs=dict(
                a=a,
                b=b,
            ),
        )

        c = vectoradd_fn_a(a)["c"]

        return jnp.sum((c) ** 2)

    if use_jit:
        loss_fn = jax.jit(loss_fn)
   
    value_and_grad_fn = jax.value_and_grad(loss_fn)

    # note this should only different wrt to the first argument a, but it passes "b" as the vjp input
    assert value_and_grad_fn(a, b) is not None
```

with a tesseract that is defined as:
```


from typing import Any

from pydantic import BaseModel, Field
from tesseract_core.runtime import Array, Differentiable, Float32


class InputSchema(BaseModel):
    a: Differentiable[Array[(None,), Float32]] = Field(description="Arbitrary vector a")
    b: Array[(None,), Float32] = Field(description="Arbitrary vector b")


class OutputSchema(BaseModel):
    c: Differentiable[Array[(None,), Float32]] = Field(
        description="Vector s_a·a + s_b·b"
    )


def apply(inputs: InputSchema) -> OutputSchema:
    """Adds two vectors `a` and `b`."""
    return OutputSchema(
        c=inputs.a + inputs.b,
    )


def abstract_eval(abstract_inputs):
    """Abstract evaluation of the addition operation."""
    return {
        "c": abstract_inputs.a,
    }


def vector_jacobian_product(
    inputs: InputSchema,
    vjp_inputs: set[str],
    vjp_outputs: set[str],
    cotangent_vector: dict[str, Any],
):
    return {
        "a": cotangent_vector["c"],
    }

```

### Logs

```bash
...
E       ValidationError: 1 validation error for endpoint vector_jacobian_product
E       body.vjp_inputs.1
E         Input should be 'a' [type=literal_error, input_value='b', input_type=str]
E           For further information visit https://errors.pydantic.dev/2.11/v/literal_error
E       --------------------
E       For simplicity, JAX has removed its internal frames from the traceback of the following exception. Set JAX_TRACEBACK_FILTERING=off to include these.

tests/test_endtoend.py:592: JaxRuntimeError
```

### OS

Linux

### Tesseract + Tesseract-JAX version

all versions

andrinr:
#### Relevant issue or PR

Resolves #137 when merged. Note the commit history is that long because I started of #127 

#### Description of changes

All changes and specific behaviours are documented in get_started.md under sharp bits. The below comments are outdated:

This issue occurs when we JIT and requests grads of a Tesseract, where only a subset of the input arguments are not marked as differentiable. In the current state we do:

1. Mark non-traced inputs as static; hence, derivatives are not requested for such inputs.
2. We compute the `vjp_inputs `, the arguments for which we request the gradient, based on all arguments minus the static args.
3. In `_pytree_to_tesseract_flat`, if a path does not match a scheme in `schema_paths ` (the differentiable input paths), we set the value to `None`. 

Based on the above, we actually have a signal to detect the issue: that is, an argument is not marked as static but is set to `None`. Therefore, inside `tesseract_compat`, we filter `vjp_inputs` based on this condition. Subsequently, we don't request the derivatives wrt. arguments that are not marked as differentiable. However, this also means we need to return something for the gradients wrt. to inputs that are `None` because JAX expects gradients wrt. to all arguments that are not static. Hence, we simply return zero arrays for such cases.

> [!WARNING]
>  While this fixes the original bug, I am not 100% sure if this is what we want. Imagine a Tesseract user forgot to mark a path as differentiable in the input scheme; with the changes here, this can fail silently with wrong gradients. It might be beneficial to log a warning.

Note the above has only been applied to VJPs, for JVPs we already have this behavior as introduced in #127 (Hence the same warning applies here as well).  See code here: 

```
flat_tangents = _pytree_to_tesseract_flat(
    tangent_inputs, schema_paths=self.differentiable_input_paths
)
[...]
out_data = self.client.jacobian_vector_product(
    inputs=primal_inputs,
    jvp_inputs=list(flat_tangents.keys()),
    jvp_outputs=jvp_outputs,
    tangent_vector=flat_tangents,
)
```
Note the `jvp_inputs` are derived from the `flat_tangents`. And the function `_pytree_to_tesseract_flat` does remove all dict entries that are not in `differentiable_input_paths`. However this bit of code:

```
 if path in out_data:
                out.append(out_data[path])
            else:
                out.append(jax.numpy.full_like(aval, jax.numpy.nan))

```

would returns NaN in those cases, we have replaced this with zeros now.

> [!WARNING]
>  Currently we are somewhat testing for derivatives of a subset of outputs of a Tesseract. But I still have the suspicion that there could be cases that break tesseract-jax.

While working on this PR. Apart from resolving the main issue, this PR also:
- Adds a new test that triggers the exact undesired behavior. Note that it works without JIT but fails when JIT is enabled. This is because there is no way to distinguish a traced array due to jitting and a traced array due to differentiability.
- Add test to compare against raw implementations for various cases of partial derivatives and back / forward mode differentiation.
- Add comments / docstrings.
- Fix sharp bits in readme as this was adressed in earlier PR #85

### DIFF
--- a/docs/content/get-started.md
+++ b/docs/content/get-started.md
@@ -77,4 +77,4 @@ When creating a new Tesseract based on a JAX function, use `tesseract init --rec
 ```
 
 
-- **Non-differentiable inputs/outputs**: Differentiating through inputs or outputs not marked as `Differentiable[...]` in the Tesseract schema can raise a `ValueError` or produces `NaN` tangents. See the [Handling Differentiability](handling-differentiability.md) page for details and workarounds.
+- **Non-differentiable inputs/outputs**: Differentiating through inputs or outputs not marked as `Differentiable[...]` in the Tesseract schema can raise a `ValueError` or produce `NaN` tangents. See the [Handling Differentiability](handling-differentiability.md) page for details and workarounds.


tesseract_jax/primitive.py
```diff
@@ -407,15 +439,7 @@ def apply_tesseract(
...
-        output_pytreedef, avals = None, None
-        avals = client.abstract_eval(
-            array_args,
-            static_args,
-            input_pytreedef,
-            output_pytreedef,
-            avals,
-            is_static_mask,
-        )
+        avals = client.abstract_eval(inputs)
```
andrinr: abstract_eval does not need to have the same signature, hence I simplified it.

tesseract_jax/tesseract_compat.py
```diff
@@ -294,7 +138,8 @@ def jacobian_vector_product(
             if path in out_data:
                 out.append(out_data[path])
             else:
-                out.append(jax.numpy.full_like(aval, jax.numpy.nan))
+                # Missing paths mean zero gradient
+                out.append(jax.numpy.zeros(aval.shape, dtype=aval.dtype))
```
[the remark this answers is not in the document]
andrinr: Good catch, I did actually forget that I have made this change as well. This change is responsible to fix the error for forward diff / vjps. I think it can cause an undesired behaviour (see first warning in PR description).

README.md
```diff
@@ -90,7 +90,7 @@ jax.grad(vector_sum)(x, y) # 🎉
...
   ```
-- **Additional required endpoints**: Tesseract-JAX requires the [`abstract_eval`](https://docs.pasteurlabs.ai/projects/tesseract-core/latest/content/api/endpoints.html#abstract-eval) Tesseract endpoint to be defined for all operations. This is because JAX mandates abstract evaluation of all operations before they are executed. Additionally, many gradient transformations like `jax.grad` require [`vector_jacobian_product`](https://docs.pasteurlabs.ai/projects/tesseract-core/latest/content/api/endpoints.html#vector-jacobian-product) to be defined.
+- **Additional required endpoints**: Tesseract-JAX requires the [`abstract_eval`](https://docs.pasteurlabs.ai/projects/tesseract-core/latest/content/api/endpoints.html#abstract-eval) Tesseract endpoint to be defined when used in conjunction with automatic differentiation and JAX transformations. This is because JAX in these cases, mandates abstract evaluation of all operations before they are executed. Additionally, many gradient transformations like `jax.grad` require [`vector_jacobian_product`](https://docs.pasteurlabs.ai/projects/tesseract-core/latest/content/api/endpoints.html#vector-jacobian-product) to be defined.
```
xalelax: maybe add this comma or remove the one afterwards

```suggestion
- **Additional required endpoints**: Tesseract-JAX requires the [`abstract_eval`](https://docs.pasteurlabs.ai/projects/tesseract-core/latest/content/api/endpoints.html#abstract-eval) Tesseract endpoint to be defined when used in conjunction with automatic differentiation and JAX transformations. This is because JAX, in these cases, mandates abstract evaluation of all operations before they are executed. Additionally, many gradient transformations like `jax.grad` require [`vector_jacobian_product`](https://docs.pasteurlabs.ai/projects/tesseract-core/latest/content/api/endpoints.html#vector-jacobian-product) to be defined.
```

tesseract_jax/primitive.py
```diff
@@ -24,6 +25,13 @@
 
 
 class _Hashable:
+    """A wrapper class to make non-hashable objects hashable by using their id.
+
+    This is not a proper hash function, as two identical objects with different memory
+    addresses will have different hashes.
+
```
xalelax:
```suggestion
```

tests/test_api.py
```diff
@@ -192,12 +285,168 @@ def f(diffable_inputs):
...
+
+    if mode == "fwd":
+        # Forward mode: JVP
+        # Note: k, m, z0, z1 get passed as tracers to apply_tesseract.
+        # The schema says they are non-differentiable, so they should not be included in jvp_inputs.
+        primal, tangent = jax.jvp(
+            loss_fn, (diffable_inputs, k, m, z0, z1), (diffable_inputs, k, m, z0, z1)
+        )
+        assert primal is not None
+        assert tangent is not None
```
xalelax: any better assertion on these guys rather than just not none? (same with value_and_grad result a few lines below)

xalelax (commented):
Good work @andrinr ! Let's have a call about the issue you highlighted in the warning box today

andrinr:
@dionhaefner 

Yes exactly, we should not merge this until we are clear there is no better way or we have verified for sure that the approaches you mentioned do not work (which is my suspicion). We have not gone over this yet, will follow up here with news :)

> so users either make them diffable or explicitly use stop_gradients

I am pretty sure I have already tried the stopping gradients trick, and it doesn't work. I have a vague idea as to why, but will dive a bit deeper and provide a better explanation as to why it doesn't.

andrinr:
Unfortunately I know for a fact that returning NaNs dosnt work, also stop gradients. The issue with arrays being traced will remain with stop grads. In fact this issue appears in specific cases where the gradients are never requested to begin with.

dionhaefner:
To solve this for more complex cases we might want to introduce symbolic zeros in Tesseract Core. For example, when requesting `dx/da` and `dz/da` but `z` doesn't depend on `a` (only `b`), there should be a way to signal that `dz/da` is exactly zero so we can stop the chain at that point. 

This is in essence the generalized form of (1): we don't need true data independence, just independence from a subset of the inputs.

tesseract_jax/primitive.py
```diff
@@ -97,18 +133,13 @@ def tesseract_dispatch_jvp_rule(
...
     tan_args_ = tuple(
-        (
-            jax.numpy.zeros_like(arg.aval)
-            if isinstance(arg, jax._src.ad_util.Zero)
-            else arg
-        )
-        for arg in tan_args
+        (jax.numpy.zeros_like(arg.aval) if not has_tan else arg)
+        for arg, has_tan in zip(tan_args, has_tangent, strict=True)
     )
```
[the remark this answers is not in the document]
jpbrodrick89: sure, that's fine :)

docs/content/get-started.md
```diff
@@ -70,8 +70,53 @@ Now you're ready to jump into our [examples](https://github.com/pasteurlabs/tess
...
 
 ```{tip}
 When creating a new Tesseract based on a JAX function, use `tesseract init --recipe jax` to define all required endpoints automatically, including `abstract_eval` and `vector_jacobian_product`.
 ```
+
+- **Non-differentiable outputs**: When an output is marked as non-differentiable in the Tesseract API, its behavior differs by differentiation mode.
+
+  In **forward mode (JVP)**, tangents for non-differentiable outputs are `NaN`, which propagates to any downstream computation that depends on them.
+
+  In **reverse mode (VJP)**, non-differentiable outputs are silently excluded from the cotangent sum. If you accidentally leave such an output in the return value and pass a cotangent that includes it, you will get a `ValueError` due to a pytree structure mismatch:
```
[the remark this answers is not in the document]
andrinr: If you assume that "f" is defined as below with the closure you cannot pop the nondiff_res from the primals as they are already excluded.

PR #177, opened 2026-04-17
fix: prevent random deadlocks by not allocating JAX memory from within callbacks

dionhaefner:
#### Relevant issue or PR

n/a

#### Description of changes

Through some very painful debugging I realized that using `jax.numpy` from within Python callbacks can lead to deadlocks (which makes intuitive sense to me, since it's kind of doing JAX-in-JAX which seems icky at best). Let's only return NumPy arrays instead.

#### Testing done

manual, deadlock resolved


andrinr (approved):

jpbrodrick89:
@dionhaefner sorry just got around to looking at this now, we stlll test jax Tesseracts with `from_tesseract_api` in `test_api.py` (through tests/pytree_tesseract/tesseract_api.py). Are they now examples of bad practice or is our warning overly conservative? Could we add a test that replicates and guards against your deadlocks so we know the specific case where deadlocks become an issue?

PR #183, opened 2026-05-07
doc: pull in new navbar from tesseract-core

dionhaefner:
#### Relevant issue or PR

https://github.com/pasteurlabs/tesseract-core/pull/562

#### Description of changes

Pull in Tesseract ecosystem navbar. Also deletes most demos since they now live in Tesseract Core.

#### Testing done

:eyes:


jpbrodrick89 (approved):
Happy with this in theory, I did not appreciate we used tesseract-jax in our tesseract-core demos, separation of concerns and making tesseract-jax lean is indeed probably the right choice. While I do know some html and css and can likely debug issues as they arise, I'm not sure how best to actually review a html/css PR. I'm assuming most of this is copy-paste from tesseract-core anyway so approving based on that and your visual inspection, let me know if you'd like me to go deeper.

Issue #188, opened 2026-05-20
VJP dispatcher silently drops missing keys instead of raising


andrinr:
### Description

**Update after attempting to reproduce: the case I described is not reachable in practice with current `tesseract-core 1.8.0`. Leaving the issue open in case the dead-code branch matters for future-proofing, but it is not a live bug.**

When I tried to build a Tesseract whose `vector_jacobian_product` returns a dict missing one of the requested `vjp_inputs` keys (in this case `y`), the call fails server-side at `tesseract_core/runtime/core.py:299`, not silently in tesseract-jax:

```
ValidationError: 1 validation error for VJPOutputSchema
  Value error, Expected keys {'x', 'y'} in output; got {'x'}
```

`tesseract_core.runtime.core.vector_jacobian_product` validates the response with `VJPOutputSchema.model_validate(out, context={"input_keys": payload.vjp_inputs})`, so a partial response is rejected before tesseract-jax's compat layer ever runs. The 500 propagates up through `apply_tesseract` and becomes a clear `RuntimeError` for the caller.

That said, the loop at `tesseract_jax/tesseract_compat.py:215-240` still has a structurally-reachable fall-through if `path not in out_data` and `has_tangent[tan_idx]` is True:

```python
for path in flat_inputs:
    if path in out_data:
        out.append(out_data[path])
        tan_idx += 1
    elif (... non-static, no-tangent ...):
        out.append(np.full(..., np.nan, ...))
        tan_idx += 1
    # else: silently fall through — no append, no tan_idx increment
```

In current configurations, the server-side validator means this branch can only be reached if someone subclasses `Jaxeract` or bypasses the runtime layer. If that constraint ever loosens (different transport, custom client), the silent-drop would re-surface. Defensive option: raise from `Jaxeract.vector_jacobian_product` if `out_data` is missing any key that the caller requested via `has_tangent`, instead of letting the loop walk off the end of the input list.

The corresponding JVP path at line 161 silently fills NaN for missing output keys regardless of validator (the JVP response schema does not enforce key presence the same way). That part is still observable behaviour; tracked separately in #187.

### Original report (kept for context)

[...]

PR #190, opened 2026-05-20
doc: correct comment on VJP non-diff input placeholder (NaN, not zero)

### LINKED ISSUE #189: Comment / code mismatch in VJP non-diff placeholder: comment says "zero", code uses NaN
### Description

At `tesseract_jax/tesseract_compat.py:225-232` the inline comment and the code disagree on what gets returned:

```python
elif (
    tan_idx < len(has_tangent)
    and not is_static_mask[all_idx]
    and not has_tangent[tan_idx]
):
    # Non-differentiable but non-static input: return zero gradient
    # with the same shape/dtype as the corresponding input array
    out.append(
        np.full(
            array_args[array_idx].shape,
            np.nan,
            dtype=array_args[array_idx].dtype,
        )
    )
```

Pick one. The JVP path at line 161 uses `np.nan` deliberately as a sentinel (tested at `tests/test_api.py:225` via `assert jnp.all(jnp.isnan(jvp["metadata"]))`), so if the VJP path means to follow the same convention the comment should say "NaN sentinel" to match. Otherwise change the value to `0`.

This branch's output ends up in a tuple slot that JAX's transpose rule doesn't read for any user-requested derivative (see #188 follow-up), so this is doc/code consistency only, not a behaviour fix.

### Steps to reproduce

n/a — read the source.

### Logs

n/a.

### OS

Linux

### Tesseract + Tesseract-JAX version

tesseract-jax `main` (verified 2026-05-20).

andrinr:
#### Relevant issue or PR

Closes #189.

#### Description of changes

The comment at `tesseract_jax/tesseract_compat.py:225-226` said *"return zero gradient"* while the code fills the slot with `np.nan`:

```python
# Non-differentiable but non-static input: return zero gradient
# with the same shape/dtype as the corresponding input array
out.append(
    np.full(array_args[array_idx].shape, np.nan, dtype=...),
)
```

Update the comment to describe the actual behaviour, add a one-liner about why the slot exists (tuple-length contract; JAX's transpose machinery doesn't consume it for any user-requested derivative — see #188's follow-up comment for the empirical check), and cross-reference the JVP path at line ~157 which uses the same NaN sentinel convention.

No behaviour change.

#### Testing done

Comment-only edit; no runtime behaviour to test. Diff is six lines added, two removed.

### DIFF
--- a/tesseract_jax/tesseract_compat.py
+++ b/tesseract_jax/tesseract_compat.py
@@ -226,8 +226,7 @@ class Jaxeract:
                 # placeholder of the same shape/dtype as the corresponding
                 # input array. The slot exists for tuple-length contract;
                 # JAX's transpose machinery doesn't consume it for any
-                # user-requested derivative. Mirrors the NaN sentinel used
-                # in the JVP path for non-differentiable outputs (line ~157).
+                # user-requested derivative.
                 out.append(
                     np.full(
                         array_args[array_idx].shape,
--- a/tesseract_jax/tesseract_compat.py
+++ b/tesseract_jax/tesseract_compat.py
@@ -222,8 +222,12 @@ class Jaxeract:
                 and not is_static_mask[all_idx]
                 and not has_tangent[tan_idx]
             ):
-                # Non-differentiable but non-static input: return zero gradient
-                # with the same shape/dtype as the corresponding input array
+                # Non-differentiable but non-static input: return a NaN
+                # placeholder of the same shape/dtype as the corresponding
+                # input array. The slot exists for tuple-length contract;
+                # JAX's transpose machinery doesn't consume it for any
+                # user-requested derivative. Mirrors the NaN sentinel used
+                # in the JVP path for non-differentiable outputs (line ~157).
                 out.append(
                     np.full(
                         array_args[array_idx].shape,


tesseract_jax/tesseract_compat.py
```diff
@@ -222,8 +222,12 @@ def vector_jacobian_product(
...
                 and not has_tangent[tan_idx]
             ):
-                # Non-differentiable but non-static input: return zero gradient
-                # with the same shape/dtype as the corresponding input array
+                # Non-differentiable but non-static input: return a NaN
+                # placeholder of the same shape/dtype as the corresponding
+                # input array. The slot exists for tuple-length contract;
+                # JAX's transpose machinery doesn't consume it for any
+                # user-requested derivative. Mirrors the NaN sentinel used
+                # in the JVP path for non-differentiable outputs (line ~157).
```
dionhaefner:
```suggestion
                # Non-differentiable but non-static input: return a NaN
                # placeholder of the same shape/dtype as the corresponding
                # input array. The slot exists for tuple-length contract;
                # JAX's transpose machinery doesn't consume it for any
                # user-requested derivative.
```

dionhaefner (approved):

PR #191, opened 2026-05-22
feat: use jacobian endpoint 

jpbrodrick89:
#### Relevant issue or PR

- Closes #170 and largely makes #244 unnecessary

#### Description of changes

When Jacobian endpoint is available, adds a jvp/vjp batching rule that dispatches to this when tangents batched and primals unbatched and then performs a matmul. The high memory bandwidth means this is suboptimal compared to proper tangent batch support when batching over a small batch size. However, the only case I can think of when one might typically do this is when trying to inspect a sparse Jacobian's nonzero values using coloring rules. Instead if there is demand for sparsity support we could consider supporting sparse formats (e.g. BCOO/CSR) in Tesseract's jacobian endpoints.

The main optimisation this implementation leaves on the table is flattening the jacobian to allow a single matmul rather performing n_inputs * n_outputs separate matmuls. We'd need a candidate test case to determine whether the likely additional complexity is worth it.

This unlocks many differentiable systems applications including least squares solves and root finding using Jax-based optimisers as well as sensitivity analysis. It also allows us to reason about the future viability of a hessian endpoint to support second-order constrained optimisation.

#### Testing done

Tests are perhaps a bit bloated and might need to be pared back, let me know your thoughts.

- Added spy tests to ensure jacobian endpoints were being called
- Checked that chaining works correctly
- Stress tested dtype consistency with scan
- Vectoradd jacfwd/jacrev 100x10 jacobian 14x faster and 100x100 Jacobian 140x faster.


tesseract_jax/primitive.py
```diff
@@ -374,10 +446,40 @@ def tesseract_dispatch_batching(
...
+    jac_output_paths: tuple[str, ...] | None = None,
+    jac_mode: Literal["fwd", "bwd"] = "bwd",
 ) -> Any:
     """Defines how to dispatch batch operations such as vmap (which is used by jax.jacobian)."""
     _raise_if_unimplemented(eval_func, client)
 
+    n_primals = len(is_static_mask) - sum(is_static_mask)
+
+    # When jacfwd/jacrev vmap a JVP/VJP with primals unbatched and (co)tangents batched,
+    # materialise the entire Jacobian and apply to batch with matmul
```
dionhaefner: Can we make this configurable by the user, similar to how we support different vmap methods?

tesseract_jax/primitive.py
```diff
@@ -374,10 +446,40 @@ def tesseract_dispatch_batching(
     client: Jaxeract,
     eval_func: str,
     vmap_method: VmapMethod = None,
+    jac_input_paths: tuple[str, ...] | None = None,
+    jac_output_paths: tuple[str, ...] | None = None,
```
dionhaefner, line 473: Why do we need those? Shouldn't it be clear from the tracing which inputs/outputs are being differentiated?

tesseract_jax/primitive.py
```diff
@@ -401,6 +503,181 @@ def tesseract_dispatch_batching(
...
+    # single flattened matmul (``v_flat @ J_flat.T`` + split/reshape). The
+    # current loop is clearer and XLA's fuser handles typical low-block-count
+    # cases; revisit if a real workload shows it's a bottleneck.
+    if eval_func == "jacobian_vector_product":
+        outs: list[Any] = []
+        for i, (path, v) in enumerate(output_flat.items()):
+            aval = output_avals[i]
+            if v is None:
+                # NaN matches the eye-vmap behaviour for non-diff outputs;
+                # callers may branch on isnan.
```
dionhaefner: This warrants more explanation, I can only guess what it means.

tesseract_jax/primitive.py
```diff
@@ -401,6 +503,181 @@ def tesseract_dispatch_batching(
...
+                outs.append(
+                    jnp.full((batch_size, *aval.shape), jnp.nan, dtype=aval.dtype)
+                )
+                continue
+            out_i = out_path_to_idx[path]
+            acc: Any = None
```
dionhaefner: Pretty opaque here, what is `acc`? It feels like these branches could be separate functions or so, very hard to parse this all.

tesseract_jax/primitive.py
dionhaefner, line 1: Should this feature be covered by a benchmark? Would be nice to see the speedup confirmed there.

dionhaefner:
> and largely makes #244 unnecessary

Reference doesn't exist, what should this be pointing to?

tesseract_jax/primitive.py
jpbrodrick89: I initially thought this wouldn't be useful to turn off and most cases translate to cases where sparse representations would be preferable. But seeing as the Nystrom preconditioner involves batched matvecs that is a valid use case for turning this feature off.

jpbrodrick89:
>> and largely makes #244 unnecessary
>
> Reference doesn't exist, what should this be pointing to?

I meant the tesseract-core issue: https://github.com/pasteurlabs/tesseract-core/issues/244

tesseract_jax/primitive.py
jpbrodrick89: We don't NEED these, but they are to avoid code triplication, avoid recompute and allow potential alternative direct calls to Jaxeract.jacobian from other API's (for example a lineax TesseractLinearOperator whose materialise registration calls Jaxeract.jacobian directly). The first two are the most important. We need jac_input_paths in three places: `_batched_via_jacobian`, `tesseract_dispatch_abstract_eval` (for `eval_func="jacobian"` and `Jaxeract.jacobian`. We would have to triplicate the code or call a non-zero cost helper function in three different places. If adding a couple optional keyword arguments to the signature isn't worth the tradeoff to you I can reconsider, but my current lean is towards the current draft.

tesseract_jax/primitive.py
jpbrodrick89: sorry speculative comentary here, all we're doing is maintaining consistency with historical sequential jvp/vjp approach will ammend

jpbrodrick89:
@dionhaefner This should be ready for a second pass, I put a lot of effort into trying to make the PyTree matmul code more intuitive to the reader and more maintainble. Lmk if you have any further concrete suggestions :-)

jpbrodrick89:
3.10 tests fail because `jax.tree.reduce_associative` was only introduced in 0.7.0. Should I bump the lower bound in `pyproject.toml` or just add a version based fallback?

jpbrodrick89:
Note the benchmark is fair for jacrev (only sequential ever available) but not for jacfwd. To make it more fair I could add an ellipsis dimension to vectoradd_jax to allow auto_experimental to actually expand the primal dimensions.

tesseract_jax/primitive.py
dionhaefner: My bad, I confused this with the public API. For now this is fine, but the gradual increase of parameters here doesn't seem great. We might want to think about introducing generic kwargs or a context object that we can pass around.

jpbrodrick89: Makes sense, I'm not sure how manageable this is for Jax primitives but either way its probably something left for a separate PR to iron out in isolation.

tesseract_jax/primitive.py
```diff
@@ -374,13 +455,53 @@ def tesseract_dispatch_batching(
     client: Jaxeract,
     eval_func: str,
     vmap_method: VmapMethod = None,
+    materialise_jacobian: bool | None = None,
```
dionhaefner: Should be `materialize` :us:

tesseract_jax/primitive.py
```diff
@@ -609,6 +914,33 @@ def apply_tesseract(
             be transformed according to the method.
 
             See :doc:`/content/vmap-methods` for a detailed guide.
+        materialise_jacobian: Strategy for batching (co)tangents at a single
+            primal evaluation (i.e. ``vmap(lambda t: jvp(f, (x,), (t,)))``).
```
dionhaefner: This should mention both the `vmap(jvp)` case and the `jacfwd` / `jacrev` case.

tesseract_jax/primitive.py
```diff
@@ -609,6 +914,33 @@ def apply_tesseract(
...
+            If you need access to the entire Jacobian through
+            ``jacfwd``/``jacrev`` the default of using the ``jacobian``
+            endpoint should be much more performant. However, if your Jacobian
```
dionhaefner: Wondering if we can make a more general statement. Maybe "when batching over a large number of (co)tangents, or when using `jacfwd` / `jacrev`, ..."?

tesseract_jax/primitive.py
```diff
@@ -609,6 +914,33 @@ def apply_tesseract(
...
+                not expose a ``jacobian`` endpoint.
+
+            ``False``
+                Call the ``jacobian_vector_product``/``vector_jacobian_product``
+                endpoint once per (co)tangent (``jax.lax.map``). Useful for
+                low-rank approximations and coloring methods.
+
+            ``None`` (default)
+                Auto: use ``True`` when the Tesseract has a ``jacobian``
+                endpoint, else fall back to ``False``.
```
dionhaefner, line 948: This should come earlier in the docstring and can be much more concise (e.g. as second sentence outlining what `True`/`False`/`None` mean).

tesseract_jax/primitive.py
```diff
@@ -609,6 +914,33 @@ def apply_tesseract(
...
+            primal evaluation (i.e. ``vmap(lambda t: jvp(f, (x,), (t,)))``).
+            If you need access to the entire Jacobian through
+            ``jacfwd``/``jacrev`` the default of using the ``jacobian``
+            endpoint should be much more performant. However, if your Jacobian
+            is large and you are batching over a small number of (co)tangents
+            (e.g. to perform low-rank approximations or apply coloring
+            methods) the sequential per-(co)tangent path (``False``) may be
+            preferable. Note that the ``False`` path uses the ``vmap_method``
+            dispatch table: all non-``None`` methods route batched
+            (co)tangents through the sequential fall-back.
```
dionhaefner: I don't understand this sentence, please spend a few more words on explaining the interaction with `vmap_method` or cut.

dionhaefner:
> 3.10 tests fail because `jax.tree.reduce_associative` was only introduced in 0.7.0. Should I bump the lower bound in `pyproject.toml` or just add a version based fallback?

Bumping is fine, thanks.

dionhaefner:
> Note the benchmark is fair for jacrev (only sequential ever available) but not for jacfwd. To make it more fair I could add an ellipsis dimension to vectoradd_jax to allow auto_experimental to actually expand the primal dimensions.

Whatever you think is best :)

jpbrodrick89:
Note JAX dropped support for Python 3.10 in JAX 0.7.0 in accordance with [SPEC-0](https://scientific-python.org/specs/spec-0000/). Are you happy to follow the scientific ecosystem and drop support for Python 3.10 now (almost two years after SPEC-0 would suggest) or shall we just do version based handling until full 3.10 EOL at end October?

tesseract_jax/primitive.py
jpbrodrick89: I think I will cut this a bit. But for your understanding, for batching jvp's we just need to broadcasting primals to match tangents but for batching vjp's we can't "batch outputs" to match cotangents so we are forced to fall-back to sequential even if we've selected a different vmap method.

dionhaefner (approved):

dionhaefner:
Yea we can drop 3.10, will do the same across other repos soon.

jpbrodrick89:
With fair benchmarks we still get 20% speedup with jacfwd. Will merge as it and explore providing further optimisations as a follow-up.

PR #205, opened 2026-06-22
fix: replace deprecated not_mapped with its historical None alias

jpbrodrick89:
#### Relevant issue or PR

- #203 

#### Description of changes

JAX removed `jax.interpreters.not_mapped` as part of v0.10.2. This was always ever just an alias for `None`, so this PR just inlines that.

@dionhaefner we missed this because our `pyproject.toml` silences all deprecation warnings. Should we surface them as failures going forward or just be more vigilant when running CI locally/checking logs?

#### Testing done

CI


dionhaefner:
Already done in #203, so closing this :)

> @dionhaefner we missed this because our pyproject.toml silences all deprecation warnings. Should we surface them as failures going forward or just be more vigilant when running CI locally/checking logs?

To me this is CI working as intended, as we noticed the incompatibility immediately after a new JAX release was made. That said treating deprecation warnings as errors can be helpful, feel free to open a PR if you feel it provides value.

PR #206, opened 2026-06-25
chore: add ellipsis to vectoradd_jax, use jit, block until ready properly in benchmarks

jpbrodrick89:
#### Relevant issue or PR

- #191 

#### Description of changes

I realised we weren't properly jitting and using block until ready in our benchmarks, fixed here.

Also to get fair jacfwd benchmarks in #191 I need to add ellipsis to vectoradd_jax in a separate PR (as benchmarks will always run on previous version of test Tesseracts) to enable `auto_experimental` to do its magic.

#### Testing done

Benchmarks run on my machine. Let's see if this runs well in CI and what the impact is.


jpbrodrick89:
Slowdowns are likely just the effect of `block_until_ready`

dionhaefner (approved):

PR #207, opened 2026-06-25
feat(deps)!: drop support by Python 3.10, test Python 3.14 in CI, bump jax lb to 0.7.0

jpbrodrick89:
#### Relevant issue or PR

- #191 

#### Description of changes

Python 3.10 is nearing EOL (31 Oct 2026), and is almost two years past the [SPEC-0](https://scientific-python.org/specs/spec-0000/) recommended support window. JAX dropped support for Python 3.10 in version 0.7.0.  Retaining support for for Python 3.10 would require us to maintain a fallback implementation of `jax.tree.reduce_associative` which was only introduced in JAX 0.7.0.

#### Testing done

Bumped testing lower bound to 3.11


.github/workflows/run_tests.yml
```diff
@@ -13,21 +13,21 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         # NOTE: If bumping the minimum Python version here, also do it in
-        # ruff.toml, setup.py and other CI files as well.
+        # pyproject.toml, ruff.toml, and other CI files as well.
 
         # test with oldest and latest supported Python versions
-        python-version: ["3.10", "3.13"]
+        python-version: ["3.11", "3.13"]
```
dionhaefner: Can we upgrade this (and all other hard-coded 3.13 versions) to 3.14 please?

dionhaefner:
Let's be careful with braking changes (`!` suffix) since that may trigger a *major* version bump to `v1.0`, and amending that in retrospect is a pain.

dionhaefner:
My bad, `!` is correct here, otherwise we get a patch version bump in this repo. We should really unify this with the way T-Core versioning works.

dionhaefner (approved):

PR #209, opened 2026-06-25
feat: use DCE to leverage Tesseract's partial differentiation

jpbrodrick89:
#### Relevant issue or PR

- #191 

#### Description of changes

Tesseracts' `jacobian_vector_product` and `jacobian` endpoints support partial differential through `jvp_outputs` and `jac_outputs` which can reduce both the amount of computation the Tesseract does and the size of the output they return. Currently we don't leverage this in tesseract-jax because by default JAX's jvp rules are expected to produce tangents wrt all outputs. For jacobians particularly this can lead to much higher computation and memory flow than optimal. However, JAX offers a primitive ruleset for partial evaluation called `dce_rules` which allows us to leverage the fact that JAX knows which outputs are "dead" and never referenced and which are genuinely required. This PR attempts to leverage these to provide `jvp_outputs` and `jac_outputs` to the Tesseract.

@dionhaefner this is still a draft while I try get my head fully round it and neaten it up a bit, but I wanted to share here as it might affect whether we can to wait for this before the next release.

#### Testing done

Needs a lot more testing still.

### DIFF
--- a/tesseract_jax/dce.py
+++ b/tesseract_jax/dce.py
@@ -1,0 +1,136 @@
+# Copyright 2025 Pasteur Labs. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Dead-code-elimination rule for the ``tesseract_dispatch`` primitive.
+
+Forward-mode AD requests the derivative of *every* differentiable output even
+when only a few survive downstream (e.g. ``jacfwd`` of a function that returns
+one leaf of a multi-output Tesseract). JAX exposes the survivors to a primitive's
+DCE rule, letting us narrow the requested sub-block (``live_output_paths`` /
+``live_input_paths``) and drop the dead outvars so the Tesseract computes only
+what is used.
+
+Like :mod:`tesseract_jax.batching`, this module holds primitive-agnostic logic —
+it operates purely on the ``JaxprEqn`` and never imports ``tesseract_dispatch_p``;
+the rule is registered against the primitive in :mod:`tesseract_jax.primitive`.
+"""
+
+import jax.tree
+from jax._src.interpreters import partial_eval as pe
+
+try:
+    # Preferred location (jax >= ~0.4.34, required on 0.10 to avoid a deprecation
+    # warning). Falls back to ``jax.core`` on our 0.7.0 lower bound, where
+    # ``jax.extend.core.DropVar`` does not yet exist.
+    from jax.extend.core import DropVar, JaxprEqn
+except ImportError:  # pragma: no cover - exercised only on older JAX
+    from jax.core import DropVar
+    from jax.extend.core import JaxprEqn
+
+from tesseract_jax.tree_util import (
+    _pytree_to_tesseract_flat,
+    live_jvp_output_positions,
+)
+
+
+def tesseract_dispatch_dce_rule(
+    used_outputs: list[bool], eqn: JaxprEqn
+) -> tuple[list[bool], JaxprEqn | None]:
+    """Drop dead derivative outputs from a ``tesseract_dispatch`` equation.
+
+    JAX surfaces which outputs survive downstream as ``used_outputs``; we narrow
+    the requested sub-block (via ``live_output_paths`` / ``live_input_paths``) and
+    drop the dead outvars so the Tesseract computes only what is used.
+
+    Only ``jacobian`` and ``jacobian_vector_product`` carry prunable output
+    structure; ``apply`` and ``vector_jacobian_product`` defer to JAX's default
+    rule. This optimisation only kicks in when JAX runs DCE — i.e. under ``jit``
+    (any mode) and un-jitted reverse mode; un-jitted ``jacfwd`` is unaffected.
+
+    Inputs are always kept: the endpoints evaluate the full primal regardless of
+    which input columns are differentiated, so pruning ``used_inputs`` would be
+    incorrect.
+    """
+    # Effects-aware whole-equation drop (matches the un-pruned default exactly).
+    if not any(used_outputs):
+        return pe._default_dce_rule(used_outputs, eqn)
+
+    eval_func = eqn.params["eval_func"]
+    if eval_func == "jacobian":
+        return _dce_jacobian(used_outputs, eqn)
+    if eval_func == "jacobian_vector_product":
+        return _dce_jacobian_vector_product(used_outputs, eqn)
+    return pe._default_dce_rule(used_outputs, eqn)
+
+
+def _dce_jacobian(
+    used_outputs: list[bool], eqn: JaxprEqn
+) -> tuple[list[bool], JaxprEqn | None]:
+    """Prune a ``jacobian`` equation's (out x in) block grid to its live rectangle."""
+    in_paths = eqn.params.get("live_input_paths")
+    out_paths = eqn.params.get("live_output_paths")
+    if in_paths is None or out_paths is None:
+        # No explicit path layout to map ``used_outputs`` onto; keep everything.
+        return [True] * len(eqn.invars), eqn
+
+    n_in, n_out = len(in_paths), len(out_paths)
+    # Row-major layout: outvar ``i * n_in + j`` is block (out_path i, in_path j).
+    used = [[used_outputs[i * n_in + j] for j in range(n_in)] for i in range(n_out)]
+    live_out = [i for i in range(n_out) if any(used[i])]
+    live_in = [j for j in range(n_in) if any(used[i][j] for i in range(n_out))]
+
+    new_params = dict(
+        eqn.params,
+        live_output_paths=tuple(out_paths[i] for i in live_out),
+        live_input_paths=tuple(in_paths[j] for j in live_in),
+    )
+    # Emit the live rectangle in the same row-major order abstract_eval expects.
+    # Blocks inside the rectangle that are individually dead become DropVars.
+    new_outvars = [
+        eqn.outvars[i * n_in + j]
+        if used_outputs[i * n_in + j]
+        else DropVar(eqn.outvars[i * n_in + j].aval)
+        for i in live_out
+        for j in live_in
+    ]
+    return [True] * len(eqn.invars), eqn.replace(outvars=new_outvars, params=new_params)
+
+
+def _dce_jacobian_vector_product(
+    used_outputs: list[bool], eqn: JaxprEqn
+) -> tuple[list[bool], JaxprEqn | None]:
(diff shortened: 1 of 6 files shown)


jpbrodrick89:
Marked it as "ready for review" only because we don't seem to run CI on draft PR's.

tesseract_jax/dce.py
```diff
@@ -0,0 +1,182 @@
...
+Like :mod:`tesseract_jax.batching`, this module holds primitive-agnostic logic —
+it operates purely on the ``JaxprEqn`` and never imports ``tesseract_dispatch_p``;
+the rule is registered against the primitive in :mod:`tesseract_jax.primitive`.
+"""
+
+from jax._src.interpreters import partial_eval as pe
+
+try:
+    # Preferred location (jax >= ~0.4.34, required on 0.10 to avoid a deprecation
+    # warning). Falls back to ``jax.core`` on our 0.7.0 lower bound, where
```
dionhaefner: This is confusing. If 0.7.0 is the lower bound, why do we need this guard?

tesseract_jax/dce.py
```diff
@@ -0,0 +1,182 @@
...
+    # ``jax.extend.core.DropVar`` does not yet exist.
+    from jax.extend.core import DropVar, JaxprEqn
+except ImportError:  # pragma: no cover - exercised only on older JAX
+    from jax.core import DropVar
+    from jax.extend.core import JaxprEqn
+
+from tesseract_jax.tree_util import (
+    _pytree_to_tesseract_flat,
```
dionhaefner: Should be a public symbol if we cross-import it

dionhaefner: Also deserves a better name, I have no clue what this function does from its name.

tesseract_jax/primitive.py
```diff
@@ -14,19 +14,22 @@
...
 from tesseract_jax.batching import VMAP_METHOD_DISPATCH, VmapMethod
+from tesseract_jax.dce import tesseract_dispatch_dce_rule
 from tesseract_jax.dispatch_params import DispatchParams
 from tesseract_jax.tesseract_compat import Jaxeract
 from tesseract_jax.tree_util import (
     TransportArray,
     _pytree_to_tesseract_flat,
     combine_args,
     dummy_output_tree,
+    live_jvp_output_positions,
```
dionhaefner: Why is this in `tree_util`?

tesseract_jax/primitive.py
```diff
@@ -319,6 +334,30 @@ def tesseract_dispatch_transpose_rule(
...
         )
 
+    # The forward-mode DCE rule may have pruned this jvp equation's outputs down
+    # to ``live_output_paths``, in which case ``cotangent`` only carries the live
+    # leaves. Scatter them back to the full output layout (symbolic zeros for the
+    # pruned leaves) so the logic below — and the VJP it dispatches — sees the
+    # un-pruned structure. ``live_output_paths is None`` ⇒ nothing was pruned.
+    # ``live_output_paths`` is a jvp-only concept; a pruned
+    # ``vector_jacobian_product`` narrows its own ``has_tangent`` instead (scattered
+    # below), so this branch never fires on the VJP-transpose path.
```
dionhaefner: deslop

tesseract_jax/primitive.py
```diff
@@ -708,13 +747,22 @@ def _to_batched(arg: Any, ax: Any) -> Any:
...
     # ``keys()`` give the path order of the Jacobian's rows; ``values()`` give
-    # the corresponding ``tans`` / ``output_avals`` positions, dropping unrequested
-    # outputs on the reverse-mode path (signalled by ``has_cotangent``).
+    # the corresponding ``tans`` / ``output_avals`` positions. On the forward-mode
+    # (JVP) path, rows DCE has already declared dead are left out of the request
+    # entirely via ``live_output_paths``; on the reverse-mode (VJP) path, outputs
+    # whose cotangent is a symbolic zero are dropped via ``has_cotangent``.
+    # ``live_output_paths`` is only ever set on a ``jacobian_vector_product``
+    # equation; a pruned ``vector_jacobian_product`` narrows ``has_tangent`` instead.
     is_jvp = params.eval_func == "jacobian_vector_product"
```
dionhaefner: deslop

dionhaefner (approved):

tesseract_jax/primitive.py
```diff
@@ -154,11 +156,23 @@ def tesseract_dispatch_abstract_eval(
...
-    # Those have the same shape as the outputs
+    # apply / jvp emit one aval per (live) output leaf with the output's shape.
+    # ``live_output_paths`` (set by the DCE rule) prunes the jvp output tangents to
+    # those still used downstream; non-differentiable leaves are always retained
+    # and apply is never pruned.
     assert params.eval_func in ("apply", "jacobian_vector_product")
-    return tuple(
-        jax.core.ShapedArray(aval.shape, aval.dtype) for aval in params.output_avals
-    )
+    if params.eval_func == "jacobian_vector_product":
```
jpbrodrick89: Is it worth skipping when no DCE is occurring?

```suggestion
    if params.eval_func == "jacobian_vector_product" and params.live_output_paths is not None:
```

tesseract_jax/dce.py
```diff
@@ -0,0 +1,220 @@
...
+)
+
+
+def live_jvp_output_positions(
+    output_pytreedef: PyTreeDef,
+    n_outputs: int,
+    diff_output_paths: dict[str, Any],
+    live_output_paths: tuple[str, ...] | None,
+    static_output_mask: Sequence[bool] = (),
+) -> list[int]:
```
jpbrodrick89, line 47: tempting to refactor so this just takes `params: DispatchParams` as an argument as would reduce LoC elsewhere making the actual logic clearer.

docs/content/handling-differentiability.md
```diff
@@ -209,3 +209,33 @@ Differentiating with respect to a tangent / cotangent works, because a derivativ
 v = jnp.ones_like(x)
 jax.grad(lambda v: jax.jvp(f, (x,), (v,))[1])(v)  # ✅
 ```
+
+## Partial differentiation
```
jpbrodrick89: Note these docs are nice-to-know not critical to users learning the ropes, it all works automatically under the hood, its tempting to consider removing this section (its full of "warnings" but you still get the right answer) or move to some new more advanced page on how tesseract-jax works.

For a minimal change maybe just retitle to "(How) does tesseract-jax respect partial differentiation?" or similar.

I don't really feel that strongly about this, just wanted to highlight.

tesseract_jax/dce.py
```diff
@@ -0,0 +1,220 @@
...
+    ``live_output_paths is None`` means "keep everything" (the un-pruned default,
+    e.g. when DCE never ran).
+
+    Static (non-array) output leaves never enter the bind, so ``static_output_mask``
+    drops them from the layout via :func:`dummy_output_tree`; the positions returned
+    then index ``output_avals``, which holds arrays only.
+
+    This is the single source of truth shared by ``abstract_eval`` (which sizes
+    the primitive's outputs) and ``Jaxeract.jacobian_vector_product`` (which
+    assembles them); keeping them in lock-step is what makes pruning safe.
```
jpbrodrick89:
```suggestion
    """Output-leaf positions a ``jacobian_vector_product`` bind should emit.

    Positions are returned in ``output_avals`` order so that abstract_eval, the
    endpoint wrapper and the DCE rule all agree on the layout. A leaf is kept
    when it is non-differentiable (its tangent is a cheap NaN and cannot be named
    by a path) or when its differentiable path is in ``live_output_paths``.
```

tesseract_jax/dce.py
```diff
@@ -0,0 +1,220 @@
...
+
+    We AND ``used_outputs`` back into ``has_tangent`` and drop the dead outvars;
+    ``abstract_eval`` then recomputes the shorter output arity and the endpoint
+    scatters the survivors back into full primal order.
+    """
+    dispatch_params = eqn.params["params"]
+    old_has_tangent = dispatch_params.has_tangent
+
+    # Consume one ``used_outputs`` flag per live (True) slot; a live slot now dead
+    # flips to False.
```
jpbrodrick89: Makes the code harder to understand than it needs to be

```suggestion
```

tesseract_jax/dce.py
```diff
@@ -0,0 +1,220 @@
...
+
+    A VJP bind returns one array per differentiated primal (the ``has_tangent``-True
+    slots, in primal order; see the reverse branch of
+    ``tesseract_dispatch_abstract_eval``), so ``used_outputs`` lines up 1:1 with
+    those slots. A dead output means that primal's gradient is never consumed
+    downstream, e.g. ``jax.grad(f)(inputs)["a"]`` keeps only ``a``'s gradient.
+
+    We AND ``used_outputs`` back into ``has_tangent`` and drop the dead outvars;
+    ``abstract_eval`` then recomputes the shorter output arity and the endpoint
+    scatters the survivors back into full primal order.
```
jpbrodrick89: Main issue here is the brackets and using "AND" as a verb (if I want to read programmer, I'd just read the code), not sure what the best alternative is but this is my attempt.

```suggestion
    A VJP bind returns one array per differentiated primal. A dead output means
    that primal's gradient is never consumed downstream, e.g. 
    ``jax.grad(f)(inputs)["a"]``.

    ``has_tangent`` already exists to distinguish arrays that are closed over on
    the forward pass. Therefore, we can update this to also drop tangents
    recognised as dead by DCE after the reverse pass.
```

tests/test_jacobian_batching.py
```diff
@@ -436,6 +448,176 @@ def f_seq(x):
     )
 
 
+@pytest.mark.parametrize("use_jit", [False, True])
+def test_grad_restricts_vjp_inputs(
```
jpbrodrick89: this test and the next two are in the wrong file (but perhaps its just a poorly named file)

docs/content/handling-differentiability.md
```diff
@@ -209,3 +209,33 @@ Differentiating with respect to a tangent / cotangent works, because a derivativ
 v = jnp.ones_like(x)
 jax.grad(lambda v: jax.jvp(f, (x,), (v,))[1])(v)  # ✅
 ```
+
+## (How) does tesseract-jax respect partial differentiation?
```
```suggestion
## How Tesseract-JAX handles partial differentiation *(it's automatic under JIT!)*
```

tesseract_jax/dce.py
dionhaefner: Tried it and did not make the logic cleaner, because inputs here do not always come from `DispatchParams`

PR #213, opened 2026-07-05
chore: drop Python 3.11 support

dionhaefner:
Bumps the minimum supported Python to 3.12 (supported: 3.12–3.14), in line with the NumPy deprecation policy.

- `requires-python` → `>=3.12,<3.15`; ruff `target-version` → `py312`.
- Drop 3.11 from the `run_tests`, `test_examples` and `test_pip_install` matrices (including the oldest-JAX include row).

Please confirm no runtime dependency or code path still relies on 3.11 before merging.


jpbrodrick89:
Python 3.10 support already dropped, otherwise lgtm

jpbrodrick89 (approved):

PR #218, opened 2026-07-15
chore: migrate build backend from versioneer to hatch-vcs

dionhaefner:
#### Relevant issue or PR

Follows the `pasteur-python-template` migration from setuptools/versioneer to hatchling/hatch-vcs (template PR #87). `tesseract-core` and `mosaic` already run on hatchling; this brings `tesseract-jax` in line. Excluded from the automated template rollout because it is a public repo.

#### Description of changes

Switches the build backend from `setuptools` + `versioneer` to `hatchling` + `hatch-vcs`:

- `pyproject.toml`: replaces the `[build-system]`, `[tool.setuptools.*]`, and `[tool.versioneer]` blocks with the hatchling equivalents. `hatch-vcs` derives the version from git tags (same PEP 440 scheme as before), writing a generated `_version.py` at build time. The `requires` list matches the already-migrated repos (`setuptools-scm!=9.0.0` pin included).
- `__init__.py`: reads `__version__` from the generated `_version.py` directly instead of calling `versioneer`'s `get_versions()`.
- Removes the vendored `versioneer` `_version.py` (now build-generated and git-ignored), the `setup.py` shim, and `MANIFEST.in` (sdist contents are now declared via `[tool.hatch.build.targets.sdist]`). No `.pyx`/`.pxd` files exist, so the Cython lines in the old `MANIFEST.in` were vestigial.

No change to the published version scheme, package contents, or public API.

#### Testing done

Built locally with `python -m build`: sdist and wheel both produce the expected tag-derived version (`0.4.1.dev8+g…`), `_version.py` is generated into the wheel, and `examples/` + metadata land in the sdist. CI (tests, `test_pip_install`, publish dry-run) will confirm on this PR.


jpbrodrick89 (approved):
✅

PR #228, opened 2026-08-04
fix: guard transpose rule against UndefinedPrimal

jpbrodrick89:
#### Description of changes

Right now if a user did the following:

```python
def f(x):
    return apply_tesseract(univariate_tess, dict(x=x, y=y))["result"]

def jvp_fun(primals, tangents):
    return jax.jvp(f, (primals,), (tangents,))[1]

jax.linear_transpose(jvp_fun, x, x)(ct)
```

They'd get the following error:

```
ValueError: zip() argument 2 is longer than argument 1
  File "tesseract_jax/primitive.py", line 280, in tesseract_dispatch_transpose_rule
    p for p, m in zip(_flat_inputs, is_static_mask, strict=True) if not m
```

And if a dev wrote a test including this with typeguard on they'd get:

```
typeguard.TypeCheckError: item 0 of argument "args" (tuple) did not match any element in the union:
  jax.Array: is not an instance of jax.Array
  numpy.ndarray: is not an instance of numpy.ndarray
  numpy.bool: is not an instance of numpy.bool
  numpy.number: is not an instance of numpy.number
  bool: is not an instance of bool
  int: is not an instance of int
  float: is neither float or int
  complex: is neither complex, float or int
```

Both are opaque and unhelpful, this PR rewords the error message to the following:

```
ValueError: Transpose of the Tesseract primitive requires concrete primal values,
but received UndefinedPrimal. This typically happens when jax.linear_transpose is
applied to a function whose arguments include both primals and tangents. Close over
the primals instead, e.g.:
  primals = (x,)
  jax.linear_transpose(lambda t: jax.jvp(f, primals, (t,))[1], x)
```

The cause is that transposing with respect to a primal isn't linear, so no endpoint can serve it. Plain JAX refuses the equivalent transpose too, so this is parity rather than a tesseract-jax limitation.

The transpose rule's signature is widened to `ArrayLike | ad.UndefinedPrimal` — without that, typeguard rejects the argument before the guard can run, which is why the tests would otherwise need `suppress_type_checks()`.

#### Testing done

Adds tests for `jax.linearize` and `jax.linear_transpose` against `apply_tesseract`, which had no coverage: the guard itself, the closure-over-primals workaround the message recommends, `linearize`'s tangent function matching `jax.jvp`, and `linear_transpose` of it matching `jax.vjp`. All parametrised over `use_jit`, since eager and staged dispatch go through separately registered code paths (`tesseract_dispatch_p.def_impl` vs `mlir.register_lowering`).

Both error messages above were reproduced against this PR's base (312bdf7). Reverting the guard fails exactly the test that covers it and nothing else. Full suite green.

### DIFF
--- a/tesseract_jax/primitive.py
+++ b/tesseract_jax/primitive.py
@@ -233,7 +233,7 @@ ad.primitive_jvps[tesseract_dispatch_p] = tesseract_dispatch_jvp_rule
 
 def tesseract_dispatch_transpose_rule(
     cotangent: Sequence[ArrayLike | ad.Zero],
-    *args: ArrayLike,
+    *args: ArrayLike | ad.UndefinedPrimal,
     static_args: tuple[_Hashable, ...],
     input_pytreedef: PyTreeDef,
     output_pytreedef: PyTreeDef,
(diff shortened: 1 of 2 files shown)


dionhaefner:
@jpbrodrick89 Can you clarify the PR body a bit? I don't understand what's being fixed here, and the mention of typeguard (which isn't enabled at runtime) is confusing.

jpbrodrick89:
This is not really a "fix" to behaviour but simply providing a more helpful error message to users and add testing of a previously untested oath. Yes, typeguard off is the user experience, whuch reports a zip error, the typeguard on error is the dev test experience. Both errors are obscure and unhelpful, this new error explains to a user exactly what they might have done wrong.

Happy to prefix with docs/test instead of fix or just update the body with the an explanation worded more like the above.

dionhaefner:
We could start with a snippet that reproduces the problem? :)

jpbrodrick89:
Done

tesseract_jax/primitive.py
```diff
@@ -251,6 +251,23 @@ def tesseract_dispatch_transpose_rule(
...
+    # UndefinedPrimal fall through into the checks below.
+    if any(ad.is_undefined_primal(p) for p in primal_args):
+        raise ValueError(
+            "Transpose of the Tesseract primitive requires concrete primal "
+            "values, but received UndefinedPrimal. This typically happens when "
+            "jax.linear_transpose is applied to a function whose arguments "
+            "include both primals and tangents. Close over the primals instead, "
+            "e.g.:\n"
+            "  primals = (x,)\n"
+            "  def jvp_tangent_only(t): return jax.jvp(f, primals, (t,))[1]\n"
```
dionhaefner: I'd write this with a lambda to avoid the weird one line def but ok

dionhaefner (approved):

PR #229, opened 2026-08-06
feat: support differentiating Tesseract derivative endpoints wrt (co)tangents

jpbrodrick89:
#### Description of changes

`jacobian_vector_product` and `vector_jacobian_product` are linear wrt their provided (co)tangent vectors. This means that it is never _necessary_ to differentiate these endpoints wrt their (co)tangent vectors (one can just evaluate at the new (co)tangent vector instead and get the right answer), but some packages do this anyway, largely for the sake of convenience rather than performance. For example, optimistix's `lin_to_grad` [helper](https://github.com/patrick-kidger/optimistix/blob/6415ab15480ebdb7386e17850105ce0dba6786ec/optimistix/_misc.py#L245-L258) uses the following pattern to create a gradient from a linear function with no transpose rule:

```python
f_eval, lin_fn, aux_eval = jax.linearize(lambda _y: fn(_y, args), state.y_eval, has_aux=True)
jax.jacfwd(lin_fn)(y_eval)
```

However, if `fn` involves a Tesseract, tesseract-jax will raise a `RuntimeError: Cannot take higher-order derivatives`. By telling tesseract-jax that `jacobian_vector_product` and `vector_jacobian_product` are linear wrt their (co)tangents we can relax this and just tell JAX to re-evaluate the endpoint with the batched tangent vectors. This is exactly what this PR does, and moreover in this specific case `jacobian_vector_product` is NEVER evaluated (given `materialise_jacobian=True`), we only need one `apply` call and one `jacobian` call, JAX DCE's the "primal" `jacobian_vector_product` automatically.

#### Testing done

Added tests to ensure the desired behaviour is provided including monitoring endpoint invocations to track DCE.

### DIFF
--- a/tesseract_jax/primitive.py
+++ b/tesseract_jax/primitive.py
@@ -46,6 +46,14 @@ class _Hashable:
             return id(self.wrapped)
 
 
+def _instantiate_zeros(tangents: Sequence[Any]) -> tuple[ArrayLike, ...]:
+    """Densify symbolic zeros, which cannot be passed to ``bind`` as-is."""
+    return tuple(
+        jax.numpy.zeros_like(t.aval) if isinstance(t, jax._src.ad_util.Zero) else t
+        for t in tangents
+    )
+
+
 @tesseract_dispatch_p.def_abstract_eval
 def tesseract_dispatch_abstract_eval(
     *array_args: ArrayLike | ShapedArray,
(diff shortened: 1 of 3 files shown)


dionhaefner (approved):
lgtm, thanks

PR #233, opened 2026-08-10
perf: enable CSE with `has_side_effect=False`

jpbrodrick89:
#### Description of changes

* Sets `has_side_effect=False` unconditionally with no opt-out offered
* Add `__eq__` and `__hash__` methods on to `Jaxeract` to establish equality based on Tesseract equality (at the moment this is just identity; we could expand tesseract-core to include identity based on url but I don't think its really necessary).

The current approach of setting `has_side_effect=True` prevents XLA from being overly aggressive with CSE but does not guarantee every coded up call will result in an endpoint request (for example, DCE is still possible if no output is used at all). Given `has_side_effect=True` doesn't really do what it says on the tin (we have to look at post-DCE jaxpr to estimate the number of actual requests) and the kind of side effects that Tesseracts are expected to have (e.g. mlflow logging or other diagnostic information) should not cause issues if an endpoint call is missed I'm proposing to set `has_side_effect=False` for the reasons listed below. Behaviour will change if a Tesseract is not a pure function (i.e. there is a counter that returns a traced array based on how many times an endpoint has been called) but using JAX in such a situation would never have been safe anyway.

One particular use case is [chunking a jacobian](https://github.com/unalmis/adv-jax-math) to cap peak memory when a Tesseract is one component of a larger function: split the identity matrix, run the linearized function over each chunk, concatenate. If the Tesseract has `materialize_jacobian=True` this will result in multiple `jacobian` requests, but the Jacobian does not depend on the tangents, so every chunk issues a byte-for-byte identical output if CSE is disabled.

Note the **eager** path does not go through lowering, so an un-jitted chunk loop still issues one request per chunk.

#### Exceptions, and what https://github.com/pasteurlabs/tesseract-jax/pull/44 was protecting 

PR https://github.com/pasteurlabs/tesseract-jax/pull/44 set `has_side_effect=True` deliberately, because our callbacks can raise and the JAX docs warn that exceptions from a pure callback leave the program's behaviour undefined. Measuring this:

                                  =True              =False
    output used                   JaxRuntimeError    JaxRuntimeError
    two identical calls (CSE)     JaxRuntimeError    JaxRuntimeError
    output discarded              no error, 0 calls  no error, 0 calls
    consumer folded away in HLO   JaxRuntimeError    no error, 0 calls

CSE itself is error-safe. Folding two identical calls into one still raises, with the same exception, so the optimization this commit enables does not lose failures. https://github.com/pasteurlabs/tesseract-jax/pull/44's concern does not bear on it.

Importantly, a call whose output goes unused already issues no request and reports no error even under `has_side_effect=True`, because JAX's own DCE removes the eqn before lowering, which is where the flag lives.

One case does regress: when the output is consumed at the jaxpr level but XLA can fold the consumer away -- `jnp.where(jnp.array(False), c, a)` -- `True` keeps the call and surfaces the error where `False` elides both. Read against the row above, that makes `True` incoherent rather than protective: eliding via DCE loses the error, eliding via constant folding keeps it, and which you get depends on whichever layer happens to notice the result is dead. `False` is at least consistent -- if nobody needs the result, the Tesseract is not called and its errors are not reported.

#### Testing done

Added explicit tests to ensure expected behaviour through CI. Checked for absence of over-eager folding: calls that differ in their inputs still issue separate requests.

### DIFF
--- a/tesseract_jax/primitive.py
+++ b/tesseract_jax/primitive.py
@@ -475,6 +475,16 @@ def tesseract_dispatch_lowering(
             out = (out,)
         return out
 
+    # A Tesseract endpoint is a pure function of its inputs, so declare it as one.
+    # This is what lets XLA's CSE fold repeated identical calls into a single
+    # request -- the same treatment a LAPACK custom call such as ``lu_factor``
+    # already gets. It relies on the bind params comparing equal for identical
+    # calls, which is why ``Jaxeract`` defines ``__eq__`` / ``__hash__``.
+    #
+    # Marking the callback side-effecting would *not* buy a guarantee that it
+    # always runs: JAX's own DCE already drops a Tesseract call whose outputs are
+    # unused, before XLA ever sees it. All it did was suppress CSE and pin the
+    # call's order against other effects.
     result, _, keepalive = mlir.emit_python_callback(
         ctx,
         _dispatch,
@@ -482,7 +492,7 @@ def tesseract_dispatch_lowering(
         array_args,
         ctx.avals_in,
         ctx.avals_out,
-        has_side_effect=True,
+        has_side_effect=False,
     )
     ctx.module_context.add_keepalive(keepalive)
     return result
--- a/tesseract_jax/tesseract_compat.py
+++ b/tesseract_jax/tesseract_compat.py
@@ -55,6 +55,24 @@ class Jaxeract:
 
         self.available_methods = self.client.available_endpoints
 
+    # Every attribute above is derived from ``self.client``, so two wrappers around
+    # the same Tesseract are interchangeable. Saying so matters: this object is a
+    # parameter of ``tesseract_dispatch_p``, and ``apply_tesseract`` builds a fresh
+    # one per call, so with the inherited identity semantics two otherwise identical
+    # calls lower to custom calls that XLA cannot recognise as equal and therefore
+    # cannot common up. Equality is delegated rather than based on the URL or schema
+    # so that distinct Tesseracts stay distinct; if ``Tesseract`` ever gains value
+    # semantics of its own, this inherits them.
+    def __eq__(self, other: object) -> bool:
+        """Whether ``other`` wraps the same Tesseract."""
+        if not isinstance(other, Jaxeract):
+            return NotImplemented
+        return self.client == other.client
+
+    def __hash__(self) -> int:
+        """Hash consistently with ``__eq__``."""
+        return hash((Jaxeract, self.client))
+
     # The abstract_eval method is never called from a dispatch function,
     # hence its signature does not need to be identical to the one of apply,
     # vjp and vjp.
--- a/docs/content/get-started.md
+++ b/docs/content/get-started.md
@@ -101,16 +101,18 @@ When creating a new Tesseract based on a JAX function, use `tesseract init --rec
 - **Tesseracts are assumed pure functions** of their inputs. Tesseract-JAX lowers each
   endpoint call as a pure operation, which is what allows repeated identical calls to be
   collapsed into a single request. Where purity does not hold, the compiler is free to
-  surprise you:
+  surprise you. All of the following require compilation, so they apply under
+  `jax.jit` (including anything jitted internally) and never to eager execution, where
+  each call runs as soon as it is reached:
   - **A call whose result is provably unused may not happen.**
 
     ```python
     @jax.jit
     def unused_result(a):
-        _ = apply_tesseract(tess, inputs)["c"]   # not called: nothing depends on it
+        _ = apply_tesseract(tess, inputs)["c"]  # not called: nothing depends on it
         return a * 2.0
 
-    threshold_ok = False   # a concrete value, not a traced argument
+    threshold_ok = False  # a concrete value, not a traced argument
 
     @jax.jit
     def dead_branch():
@@ -118,10 +120,10 @@ When creating a new Tesseract based on a JAX function, use `tesseract init --rec
         return jnp.where(threshold_ok, apply_tesseract(tess, inputs)["c"], 0.0)
     ```
 
-    If the endpoint has an observable side effect — writing a file, logging to a
-    tracking server — that side effect will not happen either, and neither will any
-    error it would have raised. `abstract_eval` is still called while tracing, so a
-    Tesseract that fails abstract validation still fails.
+    If the endpoint has an observable side effect (writing a file, logging to a tracking
+    server), that side effect will not happen either, and neither will any error it
+    would have raised. `abstract_eval` is still called while tracing, so a Tesseract
+    that fails abstract validation still fails.
 
     This cuts both ways: guarding a call you know would be rejected is a legitimate way
     to avoid it, as long as the guard is something the compiler can evaluate. A guard on
@@ -130,9 +132,9 @@ When creating a new Tesseract based on a JAX function, use `tesseract init --rec
   - **How many times a call happens is not guaranteed.** Two identical calls in one
     traced function may be collapsed into one, and the compiler is in principle free to
     recompute a call to save memory. An endpoint that returns different results for
-    identical inputs — sampling without a seed input, reading mutable external state —
-    can therefore be called once where you expected twice, with both results being the
-    same value.
+    identical inputs, such as one sampling without a seed input or reading mutable
+    external state, can therefore be called once where you expected twice, with both
+    results being the same value.
 
   - **Ordering is not guaranteed** relative to other host callbacks such as
     `jax.debug.print`.
--- a/docs/content/get-started.md
+++ b/docs/content/get-started.md
@@ -97,3 +97,47 @@ When creating a new Tesseract based on a JAX function, use `tesseract init --rec
   ```{note}
   This only affects `from_tesseract_api` (in-process execution). Tesseracts served via Docker (`from_image`) run in a separate process and are not subject to this restriction.
   ```
+
+- **Tesseracts are assumed pure functions** of their inputs. Tesseract-JAX lowers each
+  endpoint call as a pure operation, which is what allows repeated identical calls to be
+  collapsed into a single request. Where purity does not hold, the compiler is free to
+  surprise you:
+  - **A call whose result is provably unused may not happen.**
+
+    ```python
+    @jax.jit
+    def unused_result(a):
+        _ = apply_tesseract(tess, inputs)["c"]   # not called: nothing depends on it
+        return a * 2.0
+
+    threshold_ok = False   # a concrete value, not a traced argument
+
+    @jax.jit
+    def dead_branch():
+        # the predicate is known at compile time, so the branch is dead
+        return jnp.where(threshold_ok, apply_tesseract(tess, inputs)["c"], 0.0)
+    ```
+
+    If the endpoint has an observable side effect — writing a file, logging to a
+    tracking server — that side effect will not happen either, and neither will any
+    error it would have raised. `abstract_eval` is still called while tracing, so a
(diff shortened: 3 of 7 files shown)


dionhaefner:
I don't see how this is addressed anywhere here except empirically:

> PR #44 set has_side_effect=True deliberately, because our callbacks can raise and the JAX docs warn that exceptions from a pure callback leave the program's behaviour undefined.

So far I'm not seeing the great upside here that warrants shouldering this risk.

jpbrodrick89:
> I don't see how this is addressed anywhere here except empirically:

The point is that when trying to come up with examples where it made a difference, only one out of four cases actually made a difference and this PR actually makes the behaviour MORE consistent with the other cases.

I can't think of any real risk to doing this, am happy to own the risk and I just think the JAX docs are being overly cautious. Tesseracts mainly raise Validation and HTTP/connection errors (in Tesseract-JAX we raise a few exception BEFORE the pure_callback). The three things the `has_side_effect=False` changes are:

1. Allow CSE when an endpoint is called with EXACTLY the same inputs by HLO value identity
2. Allow eliding of `jnp.where` (or other conditional) branches when the conditional is a closure constant (a concrete array not a tracer)
3. Allow duplication of calls when memory is tight

(Remember jaxpr-level DCE already happens when ALL outputs are unused so a Tesseract call whose outputs go unused already reports no error under has_side_effect=True, because JAX's DCE removes the equation before lowering. The guarantee this flag appears to give does not currently exist.)

1 shouldn't be a problem, the error would be raised on the first call so CSE'ing the second makes no difference on error raising. 

2 I think is the main place we have to use our brains, yes this could cause issues if someone has a workflow where they do `jnp.where(switch, apply_tesseract, pure_jax)` and they don't get any errors they configured apply_tesseract wrong until they flick the switch but I don't see how that's a problem or surprising, if someone used a simple `if` statement that is what they would expect. Similarly if an output was unused we already get computation elided so there is already no guarantee that every written `apply_tesseract` results in a request so its not something we could ever guarantee, in the PR we just accept this fact.

I tried to force 3 with various contrived cases but didn't manage, and don't think there are any obvious triggers for this, but don't see why it would be an issue.

The motivating factor is the pay-off is huge. For example when chunking jacobians when a Tesseract is part of a larger funcitons we reduce the number of identical endpoint calls (large memory throughput + latency) from the number of chunks to just one. Jacobian chunking is used ubiquitously within DESC and if we want to prevent Tesseracts causing massive unecessary slowdowns in such workflows this PR is mandatory unless we want to reimplement CSE from scratch with lu caches which would mostly amount to the same thing anyway.

Note that all JAX lapack ffi calls are essentially `has_side_effect=False` (calling `lu_factor` multiple times on the same matrix (by identity) already only result in one lapack call, I think we should mimic this in Tesseract.

dionhaefner:
> The point is that when trying to come up with examples where it made a difference, only one out of four cases actually made a difference and this PR actually makes the behaviour MORE consistent with the other cases.

That's what I mean by empirically. You verify that JAX happens to tolerate this today, which isn't the same as "it's safe".

> The three things the `has_side_effect=False` changes are:
> 
>     1. Allow CSE when an endpoint is called with EXACTLY the same inputs by HLO value identity
> 
>     2. Allow eliding of `jnp.where` (or other conditional) branches when the conditional is a closure constant (a concrete array not a tracer)
> 
>     3. Allow duplication of calls when memory is tight

How about (4) allow XLA to crash however violently it wants to if the wrapped code throws a Python exception?
 
> (Remember jaxpr-level DCE already happens when ALL outputs are unused so a Tesseract call whose outputs go unused already reports no error under has_side_effect=True, because JAX's DCE removes the equation before lowering. The guarantee this flag appears to give does not currently exist.)

Since we use effectful tokens I don't think this is true. JAX should currently never eliminate a Tesseract dispatch.

> The motivating factor is the pay-off is huge. For example when chunking jacobians when a Tesseract is part of a larger funcitons we reduce the number of identical endpoint calls (large memory throughput + latency) from the number of chunks to just one.

Did you confirm that XLA is actually smart enough to exploit this?

I'd also challenge "the pay-off is huge". We're weighing huge payoff for a niche audience vs. (slight?) UX degradation for *everyone*, which we should not take lightly.

jpbrodrick89:
> You verify that JAX happens to tolerate this today, which isn't the same as "it's safe".

  I have now added some regression tests in `test_exceptions.py` checking for clean failure messages under jit that should catch issues that arise when we update JAX. I've tried to cover a small variety of obvious things that can go wrong, but do let me know if you think we should be more comprehensive. This also pointed out an interesting "bug" in tesseract-core, if an `InputSchema` has a value-based check (e.g. bounds) and the `abstract_eval` endpoint receives abstract ShapeDType's which raises the following error:

  ```python
  TypeError: '>' not supported between instances of 'ShapeDType' and 'float'
    x = ShapeDType(shape=(), dtype='float64'), gt = 0.0
    pydantic/_internal/_validators.py:277, in greater_than_validator
  ```

  Do we want to support this properly? (I _think_ so?) Happy to submit a PR to tesseract-core to fix, lmk if its important to you to have it in before this PR (so we can test against it to understand the dynamics).

  > Since we use effectful tokens I don't think this is true.

  There's no ordered effect. emit_python_callback does return a token (returns_token defaults True) but we discard it and register no effects on the primitive, so nothing orders against it — jax.make_jaxpr(f)(x).effects is set().

  > JAX should currently never eliminate a Tesseract dispatch.

  Not true, for example consider the following:

  ```python
  @jax.jit
  def discard(a):
      _unused = apply_tesseract(tess, dict(a=a, b=b))["c"]
      return a * 2.0
  ```

  Today with `has_side_effect=True` this call never reaches HLO — DCE removes it during lowering (pe.dce_jaxpr shows DCE is what drops it). So has_side_effect, an attribute on an HLO custom call, has no custom call to be an attribute of.

  ```
  make_jaxpr(f)(x).jaxpr                            tesseract eqns = 1
  pe.dce_jaxpr(that, [True]*n_out)                  tesseract eqns = 0
  jit(f).trace(x).jaxpr                             tesseract eqns = 1
  jit(f).lower(x).as_text()        [StableHLO]      py callbacks   = 0
  jit(f).lower(x).compile().as_text()  [opt HLO]    py callbacks   = 0
  endpoint calls during tracing                     abstract_eval=1, apply=0
  endpoint calls executing                          abstract_eval=0, apply=0
  ```

  > Did you confirm that XLA is actually smart enough to exploit this?

  Yes, fully confirmed this is what is exactly what `test_chunked_jacobian_calls_endpoint_once` is designed to catch. This also applies to functions like `lambda M: tess(M)[0] for _ in range(3))` and reading off the optimised HLO to see there is only one call rather than three after this fix.

  > We're weighing huge payoff for a niche audience vs. (slight?) UX degradation for everyone

  Based on these tests I can't actually find any example where the UX is degraded in any way (all tracebacks are essentially identical before and after this PR, but note that under jit all exception get cast to `JaxRuntimeError`. Furthermore `test_still_usable_after_a_failure` was added to show that JAX can run cleanly after catching a `JaxRuntimeError`), this is what we get for a runtime-only positivity error inside the apply endpoint both before and after this PR.

  ```
  jax.errors.JaxRuntimeError: INTERNAL: CpuCallback error calling callback: Traceback (most recent call last):
    ...
    File ".venv/.../jax/_src/callback.py", line 815, in _wrapped_callback
    File "tesseract_jax/primitive.py", line 464, in _dispatch
    File "tesseract_jax/tesseract_compat.py", line 112, in apply
    File ".venv/.../tesseract_core/sdk/tesseract.py", line 925, in run_tesseract
  RuntimeError: Traceback (most recent call last):
    File ".venv/.../tesseract_core/sdk/tesseract.py", line 914, in run_tesseract
      result = self._endpoints[endpoint](parsed_payload)
    File ".venv/.../tesseract_core/runtime/core.py", line 226, in apply
      out = api_module.apply(payload.inputs)
    File "tests/validating_tesseract/tesseract_api.py", line 48, in apply
      raise ValueError(
  ValueError: x must be strictly positive, got -1.0. This depends on the value, so it cannot be caught at trace time.

  Error running Tesseract API apply: x must be strictly positive, got -1.0. ... (see above for full traceback)
  ```

  In fact I found one case where UX is arguably slightly improved. Consider the case where a Tesseract has a positivity check and raises an error, one might want to skip this with something `jnp.where(x>0, tess(x), jnp.zeros_like(x))`. With `has_side_effect=True` this was impossible, but with `has_side_effect=False` this works when x is a captured constant. This gives motivation to offer additional **opt-in** functionality for users to intentionally quash all errors (to handle the case where x is a tracer) and just full the output schema with nans in accordance with abstract eval but this is not what this PR is intending to do

dionhaefner:
Oh I'm pretty sure I added token handling at some point but must have ripped it out again. Oh well. Then it's not surprising DCE is triggered, agree.

Your bug is documented here: https://docs.pasteurlabs.ai/projects/tesseract-core/latest/content/how-to/defining-apis/#sharp-edge-abstract-eval-and-field-validators

Not sure if fixable but would be nice if so.

So the remaining risk is a potential rug-pull for users that rely on this feature now and it breaks later due to a JAX/XLA change of this undefined behavior. I don't feel great about it but will trust you that the upside is worth it.

tesseract_jax/primitive.py
```diff
@@ -475,14 +475,24 @@ def _dispatch(*args: ArrayLike) -> Any:
...
+    # A Tesseract endpoint is a pure function of its inputs, so declare it as one.
+    # This is what lets XLA's CSE fold repeated identical calls into a single
+    # request -- the same treatment a LAPACK custom call such as ``lu_factor``
+    # already gets. It relies on the bind params comparing equal for identical
+    # calls, which is why ``Jaxeract`` defines ``__eq__`` / ``__hash__``.
+    #
+    # Marking the callback side-effecting would *not* buy a guarantee that it
+    # always runs: JAX's own DCE already drops a Tesseract call whose outputs are
+    # unused, before XLA ever sees it. All it did was suppress CSE and pin the
+    # call's order against other effects.
```
dionhaefner, line 487: Can we document this as a sharp edge? I.e., if your Tesseract is stateful / wants to be run in a certain order you're out of luck (but could let us know if you need it, adding optional side effects / tokens should be no problem).

tesseract_jax/tesseract_compat.py
```diff
@@ -55,6 +55,24 @@ def __init__(self, tesseract_client: Tesseract) -> None:
...
+    # semantics of its own, this inherits them.
+    def __eq__(self, other: object) -> bool:
+        """Whether ``other`` wraps the same Tesseract."""
+        if not isinstance(other, Jaxeract):
+            return NotImplemented
+        return self.client == other.client
+
+    def __hash__(self) -> int:
+        """Hash consistently with ``__eq__``."""
+        return hash((Jaxeract, self.client))
```
dionhaefner, line 74: Why include `Jaxeract` here?

dionhaefner (approved):

tesseract_jax/tesseract_compat.py
jpbrodrick89: Otherwise `hash(jaxeract_instance) == hash(tess_instance)` which seems like a footgun, this just ensure the hashes between different object types are distinct.

tesseract_jax/primitive.py
jpbrodrick89: added, its quite verbose but I tried not to make it overly technical, would appreciate a quick scan before I merge, thanks Dion.

dionhaefner: I think it's fine (just a few too many em dashes for my taste). My only comment is that it should mention this only applies under `jit`.

PR #236, opened 2026-08-13
fix: keep list indices stable when pruning static leaves

### LINKED ISSUE #235: jax.jvp returns the wrong tangent for partially-differentiated list inputs
A `list[Differentiable[...]]` input where only some elements are differentiated gets its forward-mode tangent shipped under the wrong list index, so `jax.jvp` silently returns another element's derivative. Reverse mode is correct, so the two disagree with no error raised.

## Reproducer

A Tesseract with `w: list[Differentiable[Array[(3,), Float32]]]` and `result = 1.0*w[0] + 10.0*w[1]`, differentiating with respect to `w[1]` only, with `w[0]` passed as a concrete array:

```
jax.jvp  tangent : [1. 1. 1.]     <- this is d/dw[0]
jax.grad         : [10. 10. 10.]  <- correct
```

## Cause

`unflatten_args(..., remove_static_args=True)` replaces non-differentiated leaves with `None` and then calls `_prune_nones`, whose list branch drops them:

```python
return type(tree)(_prune_nones(v) for v in tree if v is not None)   # tree_util.py:82
```

Dropping renumbers the survivors, and `_pytree_to_tesseract_flat` derives list paths positionally:

```python
elif hasattr(elem, "idx"):
    tesseract_path += f".[{elem.idx}]"                              # tree_util.py:155-156
```

so the tangent for `w[1]` is sent to the Tesseract under the path `w.[0]`.

## Fix

Remove the filter from the list/tuple branch only, keeping it in the dict branch since dict keys are not positional. That is safe because `None` is an empty pytree node in JAX, so the leaf disappears from `tree_flatten_with_path` anyway without shifting its siblings:

```
tree_flatten_with_path([None, y])  ->  one leaf at idx 1
tree_flatten_with_path([y])        ->  one leaf at idx 0
```

With that one line changed the reproducer returns `[10. 10. 10.]`, and the existing suite is unchanged at 266 passed.

## Why it went unnoticed

The repo's own pytree fixture uses unequal shapes (`delta[0]` shape 8, `delta[1]` shape 9), so a mispathed tangent hits a shape mismatch and pydantic raises rather than returning a wrong number. Equal shapes are what make it silent.

Happy to send the PR. Draft #209 also touches `tree_util.py`, though not `_prune_nones`, so I would expect at most a textual conflict.

singhharsh1708:
Fixes #235.

`unflatten_args(..., remove_static_args=True)` replaces non-differentiated leaves with `None`, and `_pytree_to_tesseract_flat` derives list paths positionally. The list branch of `_prune_nones` dropped those `None` entries, which renumbered the survivors, so a tangent for `w[1]` was shipped under the path `w.[0]`. Reverse mode builds `vjp_inputs` from the unpruned primal tree and was unaffected, so forward and reverse disagreed with no error raised.

Keeping the `None` entries is safe because `None` is an empty pytree node in JAX. It contributes no leaf to `tree_flatten_with_path` either way, and keeping it is precisely what preserves the sibling index:

```
tree_flatten_with_path([None, y])  ->  one leaf at idx 1
tree_flatten_with_path([y])        ->  one leaf at idx 0
```

The dict branch still filters, since dict keys are not positional.

## Tests

Both fail on `main` and pass with the one-line change.

- `test_list_index_survives_static_pruning` pins the mechanism at the unit level, with no Tesseract: `unflatten_args` followed by `_pytree_to_tesseract_flat` must yield `w.[1]`.
- `test_pytree_tesseract_jvp_preserves_list_order` is end to end against the existing `pytree_tesseract` fixture, building `delta` in its original order.

That second test needs the explicit ordering. Adding `["delta.1"]` to `DIFFABLE_PATHS_PARAMS` looks like the natural regression test but is a no-op: `merge_dicts` concatenates diffable-first, so the traced array lands back at index 0, the one position where renumbering is invisible. I confirmed that a suite with `["delta.1"]` added passes unchanged against the unfixed code.

Two other shapes are also immune and are deliberately not used here: a `list` of pydantic sub-models (the dict branch rebuilds the static element as `{}`, which is not `None`, so nothing is dropped), and any list whose elements have different shapes (a mispathed tangent hits a shape mismatch and pydantic raises rather than returning a wrong number). The latter is why the existing `delta` entries, at shapes `(8,)` and `(9,)`, never surfaced this.

Full suite is 269 passed, up from 266. `ruff check` and `ruff format` clean.

## Two observations, no change made

- `_prune_nones` appears vestigial after this. Replacing the whole function body with `return tree` also passes 266/266, and the dict branch cannot affect paths either, since a `None` value contributes no leaf. I have left it alone rather than widen the change, but if you would rather drop the function entirely that is your call to make.
- `_pytree_to_tesseract_flat` annotates `schema_paths` as `dict[str, Any] | None` while the body only iterates it. That is accurate for every real caller, which passes the OpenAPI `differentiable_arrays` mapping, so nothing needs changing. Flagging it only because typeguard rejects a plain list of paths, which is what I first reached for when writing the unit test.

Happy to rebase around #209 whenever it lands. It touches `tree_util.py` but not `_prune_nones`, so I would expect at most a textual conflict.


jpbrodrick89:
Thanks for identifying, agree the prune nones is just inert ceremony, do you want to just delete the whole function then? Otherwise happy to merge this and do that as a follow up.

singhharsh1708:
Deleted it in 993e01c, so this PR now removes the function rather than fixing its list branch.

Worth noting the dict branch was equally inert, not just the list one: `None` is an empty pytree node, so it contributes no leaf whether or not it is removed, and dict paths are keyed rather than positional so they could not shift anyway.

The net change to `tree_util.py` is now twelve lines deleted. Full suite is 269 passed, and reintroducing the original filtering pruning still fails all three regression tests, so they continue to guard against it coming back.

jpbrodrick89 (approved):
Approved (this shouldn't affect performance so I will try merge before we land #240).

PR #239, opened 2026-08-17
feat: support device-to-device GPU interop

dionhaefner:
#### Relevant issue or PR

- Builds on pasteurlabs/tesseract-core#716 (lets the runtime find libcudart from pip CUDA wheels), released in tesseract-core 1.13.0. The lock now pins `tesseract-core>=1.13.0`.
- Sibling PR: pasteurlabs/tesseract-torch#42

#### Description of changes

Lets a `jax.jit`/`grad` program call a served Tesseract with its arrays kept on the GPU end to end. A call opts in by naming an on-device transport, `apply_tesseract(client, inputs, device_transport="cuda_ipc")`. Inputs that live on a CUDA device are then handed over by CUDA IPC handle and outputs come back as device arrays — no host round-trip at the Tesseract boundary. The transport is selected by name so further backends can be added without another flag.

**The default call is unchanged.** `apply_tesseract(client, inputs)` behaves exactly as before: without a transport it takes the host round-trip, so CPU-only installs and non-CUDA runs are unaffected. Under `jit` on a CUDA backend an on-device call lowers to the GPU path automatically, and falls back to the host-callback path when the native shim isn't present or the endpoint isn't supported. Opting in also requires the Tesseract be *served* with the matching `gpu_transport` (an experimental opt-in on the runtime side).

**Design choices worth reviewing:**

- **Why a C++ shim at all.** JAX has no Python-level FFI for GPU buffers — by the time data reaches a host callback it has already been copied off-device. So the CUDA lowering targets a native XLA FFI custom call (`_cuda_shim.cc`) that receives XLA's device pointers, hands them to a Python dispatch callback as `__cuda_array_interface__` views, and copies the returned device arrays back into XLA's output buffers. The CPU path keeps using the host callback.
- **Endpoint-generic dispatch.** The GPU path reuses the *same* dispatch closure the CPU lowering builds, keyed by an integer token passed to the handler as an FFI attribute. apply/jvp/vjp/jacobian all route through one handler unchanged; there's no GPU-specific endpoint logic.
- **No CUDA-library dependency.** The shim `dlopen`s the CUDA runtime at load time rather than linking it, and the transport needs no CuPy/Torch: inputs are bare `__cuda_array_interface__` views and outputs are the runtime's framework-agnostic `IpcDeviceArray`. The extension is built by a Hatch build hook and is **optional** — a build failure degrades to CPU-only when `TESSERACT_JAX_GPU_OPTIONAL=1`.
- **Conservative ordering.** The handler synchronizes XLA's stream around the dispatch (correctness-first; event-based overlap can come later), and clears the runtime's sticky error state after each call so it can't leak into JAX's next launch.

#### Testing done

New `tests/test_gpu_direct.py` covers apply, grad/jvp/vjp, materialized jacobian, non-diff inputs/outputs, and serial reuse against a served GPU Tesseract. An autouse fixture arms the FFI residency check (`TESSERACT_JAX_DEBUG_CHECK_DEVICE_PTRS`) for the whole module, so every functional test doubles as a regression guard: an accidental host round-trip on any dispatch path fails loudly instead of silently losing the GPU-direct speedup. A `pytest-benchmark` timing test tracks per-call `apply` latency as a signal. GPU CI runs the suite on CUDA 12 and 13; CPU-only paths are unaffected.


.github/workflows/publish.yml
jpbrodrick89: All three Python version are uploading to the same artifact name and colliding.

```suggestion
          name: dist-${{ matrix.python-version }}
```

jpbrodrick89:
```suggestion
          pattern: dist-*
          merge-multiple: true
```

This flattens all three artifacts into one dist.

jpbrodrick89: Unfortunately `uv build` publishes wheel as `linux_x86_64...` which gets rejected from PyPI as they expect `manylinux`... we should probably look at switching to cibuildwheel or other build system.

pyproject.toml
```diff
@@ -41,7 +42,15 @@ dev = [
...
 [build-system]
-requires = ["hatchling", "hatch-vcs", "setuptools-scm!=9.0.0"]
+requires = [
+    "hatchling",
+    "hatch-vcs",
+    "setuptools-scm!=9.0.0",
+    # Build-time deps for the native FFI shim (see hatch_build.py). jaxlib
+    # provides the XLA FFI headers; pybind11 provides the binding headers.
+    "pybind11>=2.12",
+    "jaxlib>=0.7.0",
```
jpbrodrick89: xla/ffi/api/c_api.h guarantees compatibility backwards: "XLA runtime guarantees that old API version are supported for at least 12 months", with struct_size for backwards compat. So a shim built against old headers works with newer jaxlib, not vice versa. Bumping it should probably stay a deliberate manual step, in step with dropping old JAX support.

```suggestion
    "jaxlib==0.7.*; python_version < '3.14'",
    "jaxlib==0.8.*; python_version >= '3.14'",
```

hatch_build.py
```diff
@@ -0,0 +1,141 @@
...
+
+        try:
+            self._compile(root, source, out)
+        except Exception as exc:
+            if os.environ.get("TESSERACT_JAX_GPU_OPTIONAL"):
+                self.app.display_warning(
+                    f"native FFI shim build failed ({exc}); GPU-direct dispatch "
+                    "will be unavailable. Continuing because "
+                    "TESSERACT_JAX_GPU_OPTIONAL is set."
+                )
```
jpbrodrick89: I'd consider changing the default to support source installs and just manually set the env to 0 in CI or remove/renamed the env (TESSERACT_JAX_GPU_REQUIRED) and just assert it built after.

hatch_build.py
```diff
@@ -0,0 +1,141 @@
...
+PACKAGE_DIR = Path("tesseract_jax")
+SOURCE = PACKAGE_DIR / "_cuda_shim.cc"
+
+
+class CudaShimBuildHook(BuildHookInterface):
+    """Compile the native FFI shim before the wheel is assembled."""
+
+    PLUGIN_NAME = "custom"
+
+    def initialize(self, version: str, build_data: dict) -> None:
```
jpbrodrick89, line 58: Consider skipping compilation entirely when on unsupported platforms (e.g. win32)

.github/workflows/publish.yml
```diff
@@ -18,8 +18,8 @@ jobs:
           - ubuntu-latest
           # - macos-latest
```
jpbrodrick89: We should still publish a `py3-none-any` wheel, otherwise non-Linux users will now have no wheels I believe.

tesseract_jax/primitive.py
```diff
@@ -501,7 +546,87 @@ def _dispatch(*args: ArrayLike) -> Any:
...
+    """GPU lowering: run the dispatch closure via the native FFI handler.
+
+    Falls back to the host-callback lowering if the caller did not opt into
+    ``cuda_ipc`` (``client._cuda_ipc``) or the native shim is unavailable (e.g.
+    a CPU-only install), so correctness never depends on the GPU path -- GPU
+    arrays are copied to host and back exactly as on CPU.
+    """
+    from tesseract_jax import gpu_ffi
+
+    if not client._cuda_ipc or not gpu_ffi.is_available():
```
jpbrodrick89: Are we OK with this being silent? Currently a user who explicitly passed cuda_ipc=True and has a broken/missing shim gets working-but-slow code with no signal. Should we warn when `cuda_ipc=True` and `gpu_ffi.is_available()` is False?

.github/workflows/run_tests.yml
```diff
@@ -75,9 +75,137 @@ jobs:
...
+        with:
+          enable-cache: true
+
+      # Build + install the package (this compiles the native FFI shim via the
+      # hatch build hook; the GPU job must NOT set TESSERACT_JAX_GPU_OPTIONAL, so
+      # a shim build failure fails the job).
+      - name: Install package
+        run: |
+          cp production.uv.lock uv.lock
+          uv sync --python 3.12 --extra dev --frozen
```
jpbrodrick89, line 125: Are we OK with just testing GPU on one python version? We probably want to consider JAX floor and Python version matrix?

tesseract_jax/_cuda_shim.cc
```diff
@@ -0,0 +1,478 @@
...
+  // decode to drop the non-fatal error it leaves behind (see DispatchImpl).
+  cudaError_t (*GetLastError)() = nullptr;
+};
+
+CudaRuntime& cuda_rt() {
+  static CudaRuntime rt;
+  static std::once_flag once;
+  std::call_once(once, [] {
+    const char* names[] = {"libcudart.so",    "libcudart.so.13",
+                           "libcudart.so.12", "libcudart.so.11"};
```
jpbrodrick89: Worth resolving the wheel lib dirs in gpu_ffi.py before falling back to bare sonames. Probably something to address when the tesseract-core fix lands.

tesseract_jax/_cuda_shim.cc
```diff
@@ -0,0 +1,478 @@
...
+}
+
+}  // namespace
+
+// ---------------------------------------------------------------------------
+// Python module
+// ---------------------------------------------------------------------------
+
+PYBIND11_MODULE(_cuda_shim, m) {
+  m.doc() = "Native FFI shim for GPU-direct Tesseract dispatch";
```
jpbrodrick89: the module's Python surface is three entry points marshalling ints/strings/lists, which is comfortably inside the stable ABI. Building with Py_LIMITED_API (nanobind, or pybind11 3.x) gives one cp312-abi3 wheel per platform instead of one per CPython version — 3× fewer wheels published, and new CPython releases stop requiring a release from us.

.github/workflows/publish.yml
jpbrodrick89: Separately: this matrix expansion also breaks the upload — all three legs write to name: python-package-distributions and v4+ artifacts are immutable, so legs 2 and 3 hit a name conflict, needs e.g. `name: dist-${{ matrix.python-version }}` + `pattern: dist-* / merge-multiple: true` on the download.

jpbrodrick89 (commented):
This is the first phase of my review, focussing ONLY on the build system.

There is currently no documentation on self-build (e.g. for developers to install from github rather than PyPi), this should probably details things like:
* Need for a C++17 compiler
* `TESSERACT_JAX_GPU_OPTIONAL` escape hatch.
* `--no-build-isolation` needs `hatchling, hatch-vcs, pybind11, jaxlib` pre-installed
* tags probably need to be fetched so fallback version isn't used.

tesseract_jax/tesseract_compat.py
```diff
@@ -20,13 +21,79 @@
 # WARNING: Do NOT use jax.numpy within Jaxeract methods, as they are executed from within FFI callbacks
 # and cannot safely allocate JAX arrays. Use vanilla numpy instead.
 
+# The endpoint methods are transport-agnostic: the CPU host-callback lowering
+# passes real arrays (``ArrayLike``), while the GPU FFI lowering passes bare
+# ``__cuda_array_interface__`` device views. ``Any`` admits both so a runtime
+# type-check does not reject the duck-typed GPU views.
+TransportArray = Any
```
jpbrodrick89: How about a protocol (also in tree_util) for anything that supports `shape -> tuple[int, ...]` and `dtype -> np.dtype` as properties/attributes? Note that this would not include `None` (which can sometimes be the result of `_placeholder`) but I think its better practice to type `TransportArray | None`.

tesseract_jax/primitive.py
```diff
@@ -447,22 +446,31 @@ def tesseract_dispatch_lowering(
...
         extra_kwargs["jac_output_paths"] = jac_output_paths
         extra_kwargs["jac_mode"] = jac_mode
 
-    def _dispatch(*args: ArrayLike) -> Any:
+    # ``args`` is transport-dependent: the CPU host-callback lowering passes real
+    # arrays (``ArrayLike``), while the GPU FFI lowering passes bare
+    # ``__cuda_array_interface__`` device views. Annotate as ``Any`` so this
+    # shared closure accepts both without a runtime type-check rejecting the
+    # duck-typed GPU views.
+    def dispatch(*args: Any) -> tuple:
```
jpbrodrick89: Another place to type as TransportArray when we have the protocol in place.

tesseract_jax/tesseract_compat.py
```diff
@@ -64,14 +131,65 @@ def __init__(self, tesseract_client: Tesseract) -> None:
...
-        return self.client == other.client
+        return self.client == other.client and self._cuda_ipc == other._cuda_ipc
 
     def __hash__(self) -> int:
         """Hash consistently with ``__eq__``."""
-        return hash((Jaxeract, self.client))
+        return hash((Jaxeract, self.client, self._cuda_ipc))
+
+    @contextlib.contextmanager
+    def cuda_ipc(self) -> Generator[None]:
```
jpbrodrick89, line 213: Not reviewing this as I assume it will change a lot after https://github.com/pasteurlabs/tesseract-core/pull/737 (because you won't need to modify the Tesseract output_format anymore I think, which is main janky thing about this context manager).

tesseract_jax/tesseract_compat.py
```diff
@@ -20,13 +21,79 @@
...
+    non-differentiable output. Such a slot exists only to satisfy the
+    output-tuple-length contract; JAX's transpose machinery never consumes it for
+    any user-requested derivative, so its *value* is immaterial (verified:
+    substituting any value leaves every user-visible gradient unchanged).
+
+    On the cuda_ipc path we return ``None``: the native FFI handler NaN-fills
+    XLA's (already-allocated) output buffer for that slot rather than copying from
+    a fabricated source array. On the host path we return an all-NaN array
+    directly. Either way the discarded slot is NaN, so an accidental consumer
+    surfaces loudly rather than silently -- the same poison on both transports.
```
jpbrodrick89: After #259 lands there will still be some case where these disagree:

```
┌───────────────┬───────────┬─────────────────────┬──────────────────┐
│     dtype     │ CPU (0/0) │  GPU (#239, 0xff)   │                  │
├───────────────┼───────────┼─────────────────────┼──────────────────┤
│ float16/32/64 │ nan       │ nan                 │ ✅               │
├───────────────┼───────────┼─────────────────────┼──────────────────┤
│ complex64/128 │ nan+nanj  │ nan+nanj            │ ✅ newly aligned │
├───────────────┼───────────┼─────────────────────┼──────────────────┤
│ int32 / int64 │ 0         │ -1                  │ ❌               │
├───────────────┼───────────┼─────────────────────┼──────────────────┤
│ uint8         │ 0         │ 255                 │ ❌               │
├───────────────┼───────────┼─────────────────────┼──────────────────┤
│ bool          │ False     │ 0xff (invalid PRED) │ ❌               │
└───────────────┴───────────┴─────────────────────┴──────────────────┘
```

I have provided fixes for this above by altering the memset, the main thing to adjust here is the doc string which needs to be updated and more precise.

tests/gpu_tesseract/tesseract_api.py
```diff
@@ -0,0 +1,143 @@
...
+
+class OutputSchema(BaseModel):
+    c: Differentiable[Array[(None,), Float32]] = Field(description="(a*scale + b)*mask")
+    c_sum: Array[(1,), Float32] = Field(
+        description="Non-differentiable diagnostic: sum(c), shape (1,)."
+    )
+
+
+def _to_cupy(x):
+    import cupy
```
jpbrodrick89: Why don't we import cupy at module level?

tesseract_jax/_cuda_shim.cc
```diff
@@ -0,0 +1,478 @@
...
+  // (the dispatch returned ``None``): its value is a discarded gradient/tangent
+  // that no consumer reads, so instead of copying we fill it with the byte
+  // pattern 0xff -- which is a (quiet) NaN for every IEEE float width. This keeps
+  // the device path's poison semantics identical to the host path (np.full(nan))
+  // and leaves no output buffer undefined, all without allocating a source
+  // buffer. Every other slot is a real device result we copy device->device.
+  for (size_t i = 0; i < out_descs.size(); ++i) {
+    if (result_descs[i].ptr == 0) {
+      if (cudaError_t e = rt.MemsetAsync(reinterpret_cast<void*>(out_descs[i].ptr),
+                                         0xff, out_descs[i].nbytes, stream);
```
jpbrodrick89: This fill is dtype-blind, so it produces a different diagnostic than the host path for every non-float dtype: -1 for int32, 255 for uint8, and an out-of-range 0xff PRED byte for bool (XLA's contract is 0/1).

#259 settles the host rule as "whatever 0/0 yields for the dtype" — NaN for inexact, zero otherwise. Matching it here needs no new plumbing, since describe() already reads buf.element_type() for every return buffer:

```suggestion
      // 0xff is a quiet NaN at every IEEE float width; for the other dtypes it is 
      // just a plausible value, so zero-fill them to match the host path
      const int pattern = out_descs[i].inexact ? 0xff : 0x00;
      if (cudaError_t e = rt.MemsetAsync(reinterpret_cast<void*>(out_descs[i].ptr),
                                         pattern, out_descs[i].nbytes, stream);
```

Where the `inexact` bool is set in `describe()` as commented there.

tesseract_jax/_cuda_shim.cc
```diff
@@ -0,0 +1,478 @@
...
+  }
+}
+
+// A plain description of one buffer, handed to Python.
+struct BufferDesc {
+  uintptr_t ptr;
+  std::string typestr;
+  std::vector<int64_t> shape;
+  size_t nbytes;
+};
```
jpbrodrick89, line 305:
```suggestion
struct BufferDesc {
  uintptr_t ptr;
  std::string typestr;
  std::vector<int64_t> shape;
  size_t nbytes;
  bool inexact;   // F16/F32/F64/BF16/C64/C128 -- dtypes with a NaN to spell
};
```

tesseract_jax/_cuda_shim.cc
```diff
@@ -0,0 +1,478 @@
...
+    case DT::F16:  return "<f2";
+    case DT::F32:  return "<f4";
+    case DT::F64:  return "<f8";
+    case DT::C64:  return "<c8";
+    case DT::C128: return "<c16";
+    case DT::BF16: return "<V2";  // no numpy bf16; opaque 2 bytes
+    default: return "";
+  }
+}
+
```
jpbrodrick89, line 267:
```suggestion

// Mirrors `np.issubdtype(arr.dtype, np.inexact)`
bool dtype_is_inexact(ffi::DataType dt) {
  using DT = ffi::DataType;
  switch (dt) {
    case DT::F16:
    case DT::F32:
    case DT::F64:
    case DT::BF16:
    case DT::C64:
    case DT::C128:
      return true;
    default:
      return false;
  }
}

```

Alternatively you could add this to dtype_typestr and have it return a tuple if you like.

tesseract_jax/_cuda_shim.cc
```diff
@@ -0,0 +1,478 @@
...
+  uintptr_t ptr;
+  std::string typestr;
+  std::vector<int64_t> shape;
+  size_t nbytes;
+};
+
+BufferDesc describe(ffi::AnyBuffer buf) {
+  BufferDesc d;
+  d.ptr = reinterpret_cast<uintptr_t>(buf.untyped_data());
+  d.typestr = dtype_typestr(buf.element_type());
```
jpbrodrick89, line 310:
```suggestion
  d.typestr = dtype_typestr(buf.element_type());
  d.inexact = dtype_is_inexact(buf.element_type());
```

tesseract_jax/tesseract_compat.py
```diff
@@ -20,13 +21,79 @@
...
+
+    return any(has_cuda_array_interface(v) for v in values)
+
+
+def _cast_return(value: TransportArray, *, dtype: np.dtype) -> TransportArray:
+    """Coerce a dispatch result to the return ``dtype`` without leaving the device.
+
+    On the cuda_ipc (GPU FFI) path ``value`` is a device array whose bytes the
+    FFI handler copies straight into XLA's output buffer -- it is already the
+    right dtype (the server computed it), so it is returned untouched. Wrapping
```
jpbrodrick89: This is not guaranteed, tesseract-core is deliberately not prescriptive about the output dtype of the jacobian endpoint. Nevertheless, agree the shim needs the output buffer to be consistent with the tesseract return. So this implementation of `_cast_return` is correct for now, but the docstring isn't and more importantly we need to validate what the Tesseract actually returns against the buffer in the cuda shim.

The reason this was missed is because we didn't test this on mixed dtype tesseract (we only test cuda-ipc on ONE tesseract).

tesseract_jax/_cuda_shim.cc
```diff
@@ -0,0 +1,478 @@
...
+      if (item.is_none()) {
+        rd.ptr = 0;
+        rd.nbytes = 0;
+        result_descs.push_back(rd);
+        continue;
+      }
+      py::object cai = item.attr("__cuda_array_interface__");
+      py::tuple data = cai["data"].cast<py::tuple>();
+      rd.ptr = data[0].cast<uintptr_t>();
+      rd.nbytes = out_descs[i].nbytes;  // trust XLA's expected size
```
jpbrodrick89: We need to validate here that what the Tesseract returns is what the buffer expects to avoid really hard to debug errors.

1. The jacobian endpoint does not validate dtype intentionally and could return complex64 or float64 with the same buffer size, we don't want to faff around with shimming our own cast so just raise.
2. Tesseracts only validate shapes when OutputSchema's fully constrain them, if a shape is keyed as `(None,)` with no after validator then it is possible a Tesseract returns an array of a wrong shape. On CPU this returns a callback error and fails loudly, but without the validation in the cuda shim we will fail silently.

```suggestion
      py::object cai = item.attr("__cuda_array_interface__");
      py::tuple data = cai["data"].cast<py::tuple>();
      rd.ptr = data[0].cast<uintptr_t>();
      // XLA sized this output buffer from tesseract-jax's declared avals, but
      // nothing makes the Tesseract return that dtype or shape: the jacobian
      // response schema permits any dtype, and unlike the host path we cannot
      // cast here. Compare before copying, so a mismatch is an error rather
      // than a reinterpreted result (int32 and float32 share an itemsize) or,
      // when the source is narrower, an out-of-bounds device read.
      rd.typestr = cai["typestr"].cast<std::string>();
      rd.shape = cai["shape"].cast<std::vector<int64_t>>();
      if (rd.typestr != out_descs[i].typestr || rd.shape != out_descs[i].shape) {
        return ffi::Error::InvalidArgument(
            "tesseract_jax result " + std::to_string(i) + ": Tesseract returned " +
            rd.typestr + shape_str(rd.shape) + ", expected " +
            out_descs[i].typestr + shape_str(out_descs[i].shape));
      }
      rd.nbytes = out_descs[i].nbytes;  // now known to agree
```

where shape_str is a new helper that could be added next to dtype_str

tesseract_jax/_cuda_shim.cc
```diff
@@ -0,0 +1,478 @@
...
+                                    cuda_err(e));
+      }
+      continue;
+    }
+    if (cudaError_t e = rt.MemcpyAsync(
+            reinterpret_cast<void*>(out_descs[i].ptr),
+            reinterpret_cast<void*>(result_descs[i].ptr), out_descs[i].nbytes,
+            cudaMemcpyDeviceToDevice, stream);
+        e != cudaSuccess) {
+      return ffi::Error::Internal("cudaMemcpyAsync(result) failed: " +
```
jpbrodrick89, line 523: Needs  StreamSynchronize(post) before return to avoiding freeing memory mid copy.

tesseract_jax/_cuda_shim.cc
```diff
@@ -0,0 +1,478 @@
...
+  // the device path's poison semantics identical to the host path (np.full(nan))
+  // and leaves no output buffer undefined, all without allocating a source
+  // buffer. Every other slot is a real device result we copy device->device.
+  for (size_t i = 0; i < out_descs.size(); ++i) {
+    if (result_descs[i].ptr == 0) {
+      if (cudaError_t e = rt.MemsetAsync(reinterpret_cast<void*>(out_descs[i].ptr),
+                                         0xff, out_descs[i].nbytes, stream);
+          e != cudaSuccess) {
+        return ffi::Error::Internal("cudaMemsetAsync(placeholder) failed: " +
+                                    cuda_err(e));
```
jpbrodrick89, line 514: Needs  StreamSynchronize(post) before return to avoiding freeing memory mid copy.

tesseract_jax/gpu_ffi.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def register_dispatch(fn: Callable[..., tuple]) -> int:
+    """Register a dispatch closure, returning its integer token."""
+    global _next_token
+    with _registry_lock:
+        token = _next_token
+        _next_token += 1
+        _registry[token] = fn
+    return token
```
jpbrodrick89: Every re-trace — new shapes, cache eviction, a fresh jit — adds an entry holding the gpu_dispatch closure, which holds the Jaxeract, which holds the Tesseract client and its HTTP session. A long-running service with varying shapes leaks steadily.

Since Jaxeract already has value equality (the property you rely on for XLA CSE), memoising the token on the dispatch parameters would fix the leak and reinforce the CSE folding, rather than just capping it (which is the most obvious and I think minimal viable fix).

tesseract_jax/_cuda_shim.cc
```diff
@@ -0,0 +1,478 @@
...
+          py::make_tuple(d.ptr, py::str(d.typestr), py::cast(d.shape)));
+    }
+
+    py::object out;
+    try {
+      out = cb(token, py_inputs);
+    } catch (py::error_already_set& e) {
+      return ffi::Error::Internal(std::string("dispatch callback raised: ") +
+                                  e.what());
+    }
```
jpbrodrick89, line 407: Some other places that may benefit from try catch:

```
368: py::sequence seq = py::reinterpret_borrow<py::sequence>(out);   // no type check
369: if (py::len(seq) != out_descs.size())                            // throws if not a sequence
388: py::object cai = item.attr("__cuda_array_interface__");          // throws if absent ← reachable
389: py::tuple data = cai["data"].cast<py::tuple>();
390: rd.ptr = data[0].cast<uintptr_t>();
```

Or just wrap the entire GIL block

jpbrodrick89 (commented):
Nice work, this was the most involved review I've had to do for a while! 😮‍💨 

Tests are very limited right now (just unit tests and one actual Tesseract rather than checking all edge cases in the suite still check out, this would have led to failures for mixed dtype Tesseract). Parameterising is tricky but it is possible if we want to do it

1. Make the test Tesseracts array-module agnostic with a helper picking cupy or numpy from an env var, and the bodies written against xp. _serve_tesseract already takes extra_env, which is how the GPU fixture selects its output format.
2. One parametrised fixture over ("host", "cuda_ipc") yielding the client plus the kwargs to pass, skipping the GPU leg via the existing _gpu_available().
3. Apply it to the platform-sensitive tests only — dtype handling, discarded-slot behaviour, non-differentiable inputs and outputs, jacobian fwd/bwd modes, batching.

Can discuss in our 1:1 if you like.

jpbrodrick89:
Scanned the build system personally and structure looks good, I got Claude to review the actual build system as my knowledge is not deep enough to give you a meaningful review here. The review was overwhelmingly positive. These were the highlights.

#### no aarch64

CIBW_ARCHS_LINUX: "x86_64". jaxlib publishes manylinux2014_aarch64 alongside x86_64, and arm64 is where a lot of the GPU-direct audience actually lives — GH200/GB200 are arm64, as are Graviton instances. Those users silently get the pure wheel and lose the entire feature this PR adds, with no signal beyond is_available() returning False.

cibuildwheel builds natively on ubuntu-24.04-arm now, so it's a matrix entry rather than QEMU.

#### glibc floor is stricter than jax's. 

The shim targets manylinux_2_28 while jaxlib targets manylinux2014 (glibc 2.17). On glibc 2.17–2.27 a user can run jax[cuda] fine but tesseract-jax falls back to pure. RHEL 8 is exactly 2.28 and Ubuntu 20.04 is 2.31, so most live systems are unaffected — but it's a deliberate narrowing worth a sentence in the comment, since matching jaxlib's manylinux2014 would cost little (those images ship a C++17-capable toolchain).

#### Version derivation can silently publish 0.0.0+unknown.

SETUPTOOLS_SCM_PRETEND_VERSION=$(uvx hatch version) is a nice solution to git-in-container brittleness, but if tags are ever missing the fallback version propagates straight to PyPI. A one-line assert that it doesn't start with 0.0.0 when triggered by a release would close it.

dionhaefner:
fyi, had to revert to `manylinux_2_28` because `manylinux2014` started building SciPy from source and it didn't seem worth the effort to fix it.

tesseract_jax/primitive.py
dionhaefner: Raises now

tesseract_jax/_cuda_shim.cc
dionhaefner: now discovered in Python via T-Core helper

tesseract_jax/_cuda_shim.cc
dionhaefner: Done, thanks

tesseract_jax/_cuda_shim.cc
dionhaefner: Done, thanks

tesseract_jax/gpu_ffi.py
dionhaefner: Implemented as suggested, thanks

.github/workflows/run_tests.yml
dionhaefner: Now testing with 2 versions

pyproject.toml
```diff
@@ -9,7 +9,11 @@ requires-python = ">=3.12,<4"
 # TODO: add your dependencies here *and* in requirements.txt
 dependencies = [
     "jax>=0.7.0",
-    "tesseract-core",
+    "packaging",
+    # The GPU (cuda_ipc) transport relies on the runtime CUDA package layout
+    # (``tesseract_core.runtime.cuda.{ipc,loader}``) and the ``gpu_transport``
+    # axis introduced in 1.13.0.
```
dionhaefner:
```suggestion
```

pyproject.toml
```diff
@@ -41,7 +46,26 @@ dev = [
...
 [build-system]
-requires = ["hatchling", "hatch-vcs", "setuptools-scm!=9.0.0"]
+requires = [
+    "hatchling",
+    "hatch-vcs",
+    "setuptools-scm!=9.0.0",
+    # Build-time deps for the native FFI shim (see hatch_build.py). nanobind
+    # provides the binding headers + sources and supports the CPython stable ABI
+    # (abi3), so the shim ships as one wheel per platform for all supported
+    # CPython versions. jaxlib provides the XLA FFI headers.
```
dionhaefner:
```suggestion
```

tesseract_jax/tree_util.py
```diff
@@ -1,16 +1,45 @@
...
+
+    The endpoint methods are transport-agnostic: the CPU host-callback lowering
+    passes real NumPy arrays, while the GPU FFI lowering passes bare
+    ``__cuda_array_interface__`` device views (see
+    :class:`tesseract_jax.gpu_ffi._DeviceArrayView`) and gets back the runtime's
+    ``IpcDeviceArray``. All the dispatch code reads off them is ``shape`` and
+    ``dtype``, so this protocol captures exactly that surface -- narrow enough
+    that the duck-typed GPU views satisfy it without importing a CUDA array
+    library. ``runtime_checkable`` so typeguard admits both transports at the FFI
+    boundary rather than rejecting the GPU views.
```
dionhaefner:
```suggestion
    ``dtype``, so this protocol captures exactly that surface.
```

tesseract_jax/tree_util.py
```diff
@@ -1,16 +1,45 @@
...
+
+    A discarded derivative slot can be ``None`` on the cuda_ipc path (see
+    :func:`tesseract_jax.tesseract_compat._placeholder`); annotate those sites
+    ``TransportArray | None``.
```
dionhaefner:
```suggestion
```

tesseract_jax/primitive.py
```diff
@@ -880,6 +963,16 @@ def apply_tesseract(
...
+        device_transport: Name of the on-device transport used to exchange GPU
+            arrays with the Tesseract instead of a host round-trip (currently
+            ``"cuda_ipc"``). Requires a served Tesseract (``HTTPClient``) started
+            with the matching ``gpu_transport`` in its ``runtime_config`` and a
+            GPU-backed JAX (arrays on a ``cuda`` device); has no effect on CPU
+            arrays or a local (in-process) client, which already shares memory.
+            For ``cuda_ipc`` both processes must share the CUDA IPC namespace
+            (Docker's ``--ipc=host``). When ``None`` (default), GPU arrays take
+            the same host round-trip as CPU arrays. This is an experimental
+            tesseract-core feature; see ``tesseract_core.runtime.cuda.ipc``.
```
dionhaefner:
```suggestion
            the same host round-trip as CPU arrays. This is an experimental
            tesseract-core feature.
```

.github/workflows/run_tests.yml
```diff
@@ -82,14 +82,42 @@ jobs:
...
+    # the CPU job: each leg tracks latest, and a third leg pins the declared floor
+    # (jax==0.7.0) on the CUDA 12 / Python 3.12 combination. The floor is only
+    # tested there because 0.7 predates Python 3.14 wheels, so the 3.14 leg cannot
+    # install it -- its floor is whatever 0.8 series ships for 3.14.
+    #
+    # The shim is always compiled against the jaxlib *floor* (pinned in
+    # pyproject's [build-system].requires), because XLA's FFI ABI guarantee only
+    # runs old-shim-on-newer-runtime. So the floor leg's jaxlib==0.7 runtime is
+    # the exact version the shim was built against, and the latest legs exercise
+    # that same floor-built shim against a newer jaxlib -- the supported direction.
```
jpbrodrick89, line 101: Can probably be condensed into three sentences/bullet points rather than three paragraphs.

tesseract_jax/gpu_ffi.py
```diff
@@ -0,0 +1,190 @@
...
+    ``key`` deduplicates registrations: the FFI call reads the token at
+    execution time, so the closure must outlive lowering and cannot be released.
+    Lowering the *same* dispatch repeatedly -- a re-trace on new shapes, a cache
+    eviction, a fresh ``jit`` -- would otherwise keep allocating tokens, each
+    pinning the closure (and through it the Jaxeract, its client, and the HTTP
+    session), so a long-running service with varying shapes leaks steadily.
+
+    When ``key`` is hashable and equal to a prior call's, the existing token is
+    returned rather than a new entry created, capping the registry at the number
+    of *distinct* dispatches. ``DispatchParams`` is frozen and value-equal, so
```
jpbrodrick89, line 110: A nice-to-have would be the ability to flush this on `jax.clear_caches` but not sure its worth it, solution would be something like:

```python
class _DispatchRegistry:
    """Token -> dispatch closure, cleared alongside JAX's own caches.

    Entries are reachable from compiled executables, so they cannot be freed on
    any scope we can observe. `jax.clear_caches()` is the one moment a user
    declares the executables holding them are expendable, so the registry
    joins JAX's cache registry and is cleared with them.
    """

    def __init__(self) -> None:
        self._entries: dict[int, Callable[..., tuple]] = {}
        self._lock = threading.Lock()
        self._next_token = 0

    def cache_clear(self) -> None:
        with self._lock:
            self._entries.clear()


_registry = _DispatchRegistry()

try:  # participate in jax.clear_caches(); private, so degrade quietly
    from jax._src.util import _caches
    _caches[_registry] = None
except Exception:  # pragma: no cover
    pass
```

tesseract_jax/gpu_ffi.py
```diff
@@ -0,0 +1,190 @@
...
+    ``__cuda_array_interface__`` whose bytes the handler copies into the XLA
+    output buffers.
+
+    The native shim marshals each entry across the nanobind boundary, where
+    ``shape`` arrives as a Python ``list`` rather than a ``tuple``; the sequence
+    annotations describe that faithfully (and are normalized to tuples below).
+    """
+    fn = _registry.get(token)
+    if fn is None:
+        raise RuntimeError(f"tesseract_jax gpu_ffi: unknown dispatch token {token}")
```
jpbrodrick89, line 185: could add some flavour that this might have been caused by `jax.clear_caches()` as I think thats the main failure method right now.

tesseract_jax/primitive.py
```diff
@@ -920,7 +1013,7 @@ def apply_tesseract(
             "directly through the Tesseract client instead of apply_tesseract."
         )
 
-    client = Jaxeract(tesseract_client)
+    client = Jaxeract(tesseract_client, device_transport=device_transport)
```
jpbrodrick89, line 1016: Either here or in `Jaxeract.__init__` I would consider adding a check that a cuda GPU is available (if you'd rather silent bypassing that's fine I understand, but want to make sure its a conscious decision), to fail loudly we would do something like

```python
if device_transport is not None:
    # maybe only test if any(d.platform == "gpu" for d in jax.devices()):
    try:
        jax.devices("cuda")
    except RuntimeError:
        raise RuntimeError("..." )
```

And if users want to consciously opt-out of this check they could use [jax.lax.platform_dependent](https://docs.jax.dev/en/latest/_autosummary/jax.lax.platform_dependent.html)

tesseract_jax/_cuda_shim.cc
```diff
@@ -0,0 +1,578 @@
...
+          nb::make_tuple(d.ptr, nb::str(d.typestr.c_str()), nb::cast(d.shape)));
+    }
+
+    nb::object out;
+    try {
+      out = cb(token, py_inputs);
+    } catch (nb::python_error& e) {
+      return ffi::Error::Internal(std::string("dispatch callback raised: ") +
+                                  e.what());
+    }
```
jpbrodrick89, line 405: I think you need to catch `std::exception` too.

jpbrodrick89 (approved):
Nice, nearly there now! Another quick pass on verbose comments would be helpful too.

Do we expect this change much if we address the T-core dlpack issue?

tesseract_jax/gpu_ffi.py
dionhaefner: I don't think there's any way to trigger this right now that doesn't boil down to mutating internal state. Since our registry is append-only and unbounded I don't see how we should run into this. When clearing the lowering cache the function will just get re-traced and either hit the same token again or generate a new entry, we're fine either way.

tesseract_jax/gpu_ffi.py
dionhaefner: Feels gimmicky imo. If clearing this becomes something users would like to do we could provide our own API for it instead of maximizing surprise by plumbing it into a JAX interface.

tesseract_jax/primitive.py
dionhaefner: Done, now raises during lowering

tesseract_jax/_cuda_shim.cc
dionhaefner: Done

PR #240, opened 2026-08-17
chore: allow benchmark reporting on fork PR's

jpbrodrick89:
#### Description of changes

The benchmarks CI failed on fork PR's (e.g. #236) because GitHub withholds secrets from pull_request runs on forks, so secrets.PL_PASTEURBOT_PAT_PUBLIC was empty and create-or-update-comment threw Parameter token or opts.auth is required. This splits the benchmarks job into two, one that runs on `pull_request` and uploads artifacts (including a new simple `pr_number.txt` file) but is now un-privileged and one new privileged job than runs on the base branch (triggered by `workflow_run`) and downloads the relevant artifacts to post the comment.

The PR number is now also rendered into benchmark_report.md via a new --pr-number flag, for provenance.

Note this PR won't post a benchmark comment — workflow_run workflows only fire once they're on the default branch, so the first real test is the next fork PR after merge. We'll want the same change in tesseract-core (#667 is failing there identically).

If this works we also need to add it in tesseract-core.

#### Testing done

Locally: compare_benchmarks.py output is byte-identical to main when --pr-number is omitted (diffed both the baseline and current-only paths), and all report variants render. pre-commit clean.

The fork path can't be exercised until this is on main.

### DIFF
--- a/.github/workflows/benchmark_report.yml
+++ b/.github/workflows/benchmark_report.yml
@@ -56,22 +56,39 @@ jobs:
           fi
           echo "found=true" >> "$GITHUB_OUTPUT"
 
-      # `pr_number.txt` comes from a workflow file that a fork controls, so
-      # strip it to digits rather than trusting it. This can never reach a
-      # shell as anything but an integer.
-      - name: Read PR number
+      # `pr_number.txt` is fork-controlled: on `pull_request` the producer
+      # workflow file comes from the PR head, so a fork could name any PR and
+      # make the bot comment on an unrelated thread. Two defences:
+      #   1. Strip to digits, so it can never reach a shell as anything else.
+      #   2. Confirm the PR really belongs to this run, by matching its head SHA
+      #      against `workflow_run.head_sha` from the trusted event payload.
+      # A mismatch also happens benignly when the PR is force-pushed mid-run, so
+      # it warns and skips rather than failing; the new push posts its own run.
+      - name: Resolve and verify PR number
         id: pr
         if: steps.check.outputs.found == 'true'
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          RUN_SHA: ${{ github.event.workflow_run.head_sha }}
         run: |
           number=$(tr -dc '0-9' < benchmark-results/pr_number.txt | head -c 9)
           if [ -z "$number" ]; then
             echo "::error::Artifact contained no valid PR number"
             exit 1
           fi
+
+          pr_sha=$(gh api "/repos/$REPO/pulls/$number" -q .head.sha 2>/dev/null \
+            | grep -oE '^[0-9a-f]{40}$' || true)
+          if [ "$pr_sha" != "$RUN_SHA" ]; then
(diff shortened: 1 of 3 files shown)


.github/workflows/benchmark_report.yml
```diff
@@ -0,0 +1,109 @@
...
+          echo "found=true" >> "$GITHUB_OUTPUT"
+
+      # `pr_number.txt` is fork-controlled: on `pull_request` the producer
+      # workflow file comes from the PR head, so a fork could name any PR and
+      # make the bot comment on an unrelated thread. Two defences:
+      #   1. Strip to digits, so it can never reach a shell as anything else.
+      #   2. Confirm the PR really belongs to this run, by matching its head SHA
+      #      against `workflow_run.head_sha` from the trusted event payload.
+      # A mismatch also happens benignly when the PR is force-pushed mid-run, so
+      # it warns and skips rather than failing; the new push posts its own run.
```
dionhaefner: Not sure I understand why we need this txt file. Mosaic has a very similar setup, so you could have a look there for inspiration.

jpbrodrick89: Actually, mosaic also passes the PR number through a file, publish-meta.txt with pr_number=…. But mosaic only guards that with numeric regex, with nothing tying that number to the triggering run. Since a fork owns the producer workflow file on pull_request, it can name any PR and get PasteurBot to comment there. So copying mosaic directly would have made us weaker, not stronger.

That said, mosaic did provide some inspiration for a better solution:

- `workflow_run` payloads carry the trusted unforgeable `head_repository` as well as `head_branch`
- Together they give `owner:branch`, and `GET /pulls?head=owner:branch&state=open` allowing PR identification/resolution

Which means pr_number.txt can be deleted entirely by looking the PR up from the event payload instead. The privileged workflow then takes no fork-controlled input except the report body, so the forgery risk is no longer present.

.github/workflows/benchmarks.yml
```diff
@@ -17,7 +17,6 @@ concurrency:
 
 permissions:
   contents: read
-  pull-requests: write
```
dionhaefner, line 20: I don't think we need to specify any permissions here then? The token should already have `contents: read` by default

.github/workflows/benchmarks.yml
```diff
@@ -87,9 +86,14 @@ jobs:
 
           echo "has_baseline=true" >> $GITHUB_OUTPUT
 
+      # Secrets are withheld from `pull_request` runs on fork PRs, so the comment
+      # is posted by the privileged `Benchmark Report` workflow instead. It
+      # works out which PR this is from the `workflow_run` event payload, so
+      # nothing needs to be handed over here beyond the report itself.
```
dionhaefner:
```suggestion
      # is posted by the privileged `Benchmark Report` workflow instead.
```

dionhaefner (approved):
Thanks! Please don't forget to roll out to T-Core as well

.github/workflows/benchmarks.yml
jpbrodrick89: You're right the default is read on fork PR's but not on maintainer PR's or push to main. This is a little bit safer if there is a security risk through dependencies we're unaware of.

PR #242, opened 2026-08-17
fix: carry vmap_method and the jacobian params through nested binds

### LINKED ISSUE #241: vmap(jacfwd) fails with vmap_method=None on Tesseracts that implement jacobian
`jax.vmap(jax.jacfwd(f))` fails on any Tesseract that implements the `jacobian` endpoint, with an error telling you to pass the argument you already passed:

```python
def f(x):
    return apply_tesseract(tess, {"x": x}, vmap_method="sequential")

jax.vmap(jax.jacfwd(f))(jnp.arange(3.0))
```
```
NotImplementedError: vmap/jacobian is not supported with vmap_method=None.
Pass vmap_method='sequential', 'expand_dims', 'broadcast_all', or 'auto_experimental' ...
```

`jacrev` and `hessian` go the same way. This contradicts `docs/content/vmap-methods.md:27`, which says setting `vmap_method` explicitly is what enables these.

**Cause.** The jacobian shortcut binds a nested `jacobian` call with `vmap_method=None` hard-coded (`primitive.py`, inside `_batched_via_jacobian`). An outer vmap batches that bind, `VMAP_METHOD_DISPATCH[None]` resolves to `_raise_not_implemented`, and the caller's own `vmap_method` never reaches it.

**Regression.** `_batched_via_jacobian` arrived in fbf8ebf (#191, v0.4.0), and the shortcut auto-enables whenever the endpoint exists, so this broke working code on upgrade. Checked out fbf8ebf^ to confirm: the same snippet returns `[-2.0, 400.0, 3202.0]` there and raises on current main.

## Second, related problem

`materialize_jacobian` is never forwarded to the batching strategies. The string does not appear in `batching.py` at all, so an inner re-bind reverts to the default and re-evaluates the shortcut gate. Measured on `jacfwd(vmap(f))`:

```
materialize_jacobian=None   ->  {'apply': 3, 'jacobian': 3}
materialize_jacobian=False  ->  {'apply': 3, 'jacobian': 3}
```

`False` is documented as the escape hatch for when materializing is too expensive, so ignoring it costs exactly what the caller asked to avoid.

## Note on the fix

Threading `vmap_method` alone is not sufficient. It works on a single-input Tesseract but fails on any multi-input schema, because `jac_input_paths`, `jac_output_paths` and `jac_mode` are dropped the same way, so the inner bind loses its path restriction:

```
TypeError: dot_general requires contracting dimensions to have the same shape, got (4,) and (3,).
```

All four need to travel together. PR to follow.

singhharsh1708:
Fixes #241.

`jax.vmap(jax.jacfwd(f))` failed on any Tesseract implementing the `jacobian` endpoint, with an error telling the caller to pass the `vmap_method` they had already passed. `jacrev` and `hessian` too. It is a regression from #191: the shortcut auto-enables whenever the endpoint exists, so it broke working code on upgrade. Confirmed against `fbf8ebf^`, where the same snippet returns `[-2.0, 400.0, 3202.0]`.

Four parameters were being dropped on nested binds, and they have to travel together:

- `vmap_method` was hard-coded to `None` in the shortcut's inner `jacobian` bind, so an outer vmap resolved it to the not-implemented handler.
- `materialize_jacobian`, `jac_input_paths`, `jac_output_paths` and `jac_mode` were never forwarded from `tesseract_dispatch_batching` to the batching strategies. `materialize_jacobian` did not appear in `batching.py` at all.

Threading `vmap_method` alone looks sufficient on a single-input Tesseract and is not: without the path parameters the inner bind reverts to every differentiable path and the contraction fails on any multi-input schema with `dot_general requires contracting dimensions to have the same shape`. There is a test for exactly that case.

The re-emit goes through a small helper that returns an empty dict for a non-jacobian bind, so binds that never carried these parameters are unchanged.

## Before and after

| case | before | after |
| --- | --- | --- |
| `vmap(jacfwd(f))`, univariate | `NotImplementedError` | `[-2.0, 400.0, 3202.0]`, matching pre-#191 |
| `vmap(jacfwd(f))`, multi-input pytree | `dot_general` shape error | matches the unbatched reference |
| `materialize_jacobian=False` under nesting | `{apply: 3, jacobian: 3}` | `{apply: 3}` |

That last row is the documented escape hatch for when materializing is too expensive, so taking the shortcut anyway was costing exactly what the caller asked to avoid.

## Tests

Three added to `test_jacobian_batching.py`, all failing on `main`: the univariate case against an unbatched reference, the multi-input pytree case that catches a partial fix, and an endpoint-count assertion for `materialize_jacobian=False`.

Suite is 269 passed against 266 on main, run with the repo's own pytest config. `ruff check` clean.

Independent of #236, which touches `tree_util.py` only.


tesseract_jax/primitive.py
```diff
@@ -665,7 +671,10 @@ def _to_batched(arg: Any, ax: Any) -> Any:
         has_tangent=has_tangent,
         client=client,
         eval_func="jacobian",
-        vmap_method=None,
+        # Propagated, not hard-coded: an outer vmap batches this bind too, and
+        # ``None`` routes it to the not-implemented handler, which then tells
+        # the caller to pass the vmap_method they already passed.
```
jpbrodrick89: comments unecessary here
```suggestion
```

tesseract_jax/batching.py
```diff
@@ -41,6 +41,31 @@ def _is_ellipsis_template(template: str | None, diff_paths: dict[str, Any]) -> b
     return template is not None and "shape" not in diff_paths[template]
 
 
+def _jac_bind_kwargs(
```
jpbrodrick89: I would remove this function and thread everything through unconditionally (by accepting these in the jvp and transpose rule signatures) rather than trying inference everything through `is None`

tests/test_jacobian_batching.py
jpbrodrick89, line 1: could you parameterise all new tests over the four vmap methods for safety (testing locally shows failure for non-sequential vmap methods)?

jpbrodrick89 (commented):
Thanks very much for another quick turnaround on a tricky issue! Great to have these stress-tested and ironed out.

As suggested in the inline comments, I think it's safer to drop `_jac_bind_kwargs` and thread everything through unconditionally by adding these params to the jvp and transpose rule signatures (those two are the only rules that don't already accept them) rather than rely on `is None `inference. (note #209 renames these to live_in/output_paths so that they can be used for DCE throughout, no need to worry about here though).

One thing the tests don't cover: this only works with `vmap_method="sequential"`, the only strategy that can wrap the jacobian endpoint under a batch. With the vectorized methods the PR turns the `NotImplementedError` into a silently wrong result — sending batched primals to the jacobian endpoint yields a `(batch, *out, batch, *in)` block instead of (batch, *out, *in). On batched_tesseract (ellipsis schema, so vectorizing is legal) vmap(jacfwd(f)) returns shape (2, 3, 2, 3) where (2, 3, 3) is expected, no error raised. Simplest workaround is in tesseract_dispatch_batching, just before the dispatch:

```python
    if eval_func == "jacobian":
        vmap_method = "sequential"

    batch_fn = VMAP_METHOD_DISPATCH[vmap_method]
```

For your own `jax.vmap(jax.jacfwd(...))` use case (assuming your `tesseract_api` support batching through Ellipsis, i.e. `Array[..., dtype]`), try `materialize_jacobian=False` together with a vectorized `vmap_method` which should result in just ONE jacobian_vector_product call. Both the batch axis and jacfwd's tangent-basis axis are then absorbed as leading dims, e.g. inputs={'x': (5, 3, 3)}, tangent={'x': (5, 3, 3)} for a batch of 5 over 3 inputs. Note that with sequential the same flag gives B×N calls instead, and reverse mode can't do this today (the endpoint's cotangent has to match its own apply output). Making the batching rule for `materialize_jacobian=True` reconstruct the Jacobian this way automatically is appealing but a separate discussion.

singhharsh1708:
All four addressed in ced6602, and thank you for the vectorized-methods catch. That one was mine: the previous commit turned a `NotImplementedError` into a silent mis-shape, which is worse than what it replaced. Confirmed on `batched_tesseract` before fixing, `vmap(jacfwd(f))` returning `(2, 3, 2, 3)` against an expected `(2, 3, 3)` on all three vectorized methods, no error.

Took your guard verbatim, pinning `vmap_method` to `sequential` for a jacobian bind just before the dispatch.

`_jac_bind_kwargs` is gone, and the parameters are threaded unconditionally with the jvp and transpose rules now accepting them, so nothing is inferred from `is None` any more. Comment at the jacobian bind dropped.

On parameterising the tests, I split it by what each schema can actually take. `batched_tesseract` is ellipsis-shaped, so all four methods are legal and the new agreement test covers all four: reverting the pin fails three of them and passes only `sequential`, which is the regression guard for what you found. The fixed-shape fixtures cannot accept a batch dimension, so `expand_dims` and `broadcast_all` raise a validation error there by design, and I parameterised those tests over `sequential` and `auto_experimental` rather than assert a pre-existing failure. Say the word if you would rather see the vectorized cases asserted as raising on those fixtures too.

Suite is 276 passed, ruff clean.

Noted on #209 renaming these to `live_in/output_paths`, happy to rebase onto whatever lands there.

The `materialize_jacobian=False` plus vectorized method trick is genuinely useful, thanks. One JVP call with both the batch axis and the tangent basis absorbed as leading dims is a much better shape for what I was doing than B x N sequential calls.

jpbrodrick89 (approved):

jpbrodrick89:
Thanks so much @singhharsh1708 for taking the feedback on board! Looks great now!

PR #243, opened 2026-08-18
chore: harden CI config and version resolution

dionhaefner:
A few small robustness improvements to build and CI config.

## Changes

- **Codecov gates on patch coverage.** Reports now reflect how well the current change is covered rather than fluctuations in unrelated code. Project coverage is retained as an informational (non-gating) status. If a branch-protection/ruleset check requires `codecov/project`, reroute it to `codecov/patch`.
- **Graceful version resolution.** A missing `_version.py` (e.g. importing the package without installing it first) no longer raises a cryptic `ImportError` — it warns and falls back to `0.0.0+unknown`. Likewise, `hatch-vcs` now falls back to `0.0.0+unknown` when it can't resolve a version instead of throwing `LookupError` during a build.
- **Drop the Python upper bound** (`<3.15` → `<4`). The strict cap failed loudly on each new Python release without providing real protection.
- **Pin Node for node-based pre-commit hooks** (`prettier`). Some machines ship a system Node too old for the bundled npm, which crashes hook installation with an unhelpful error; pinning provisions a known-good Node via nodeenv.
- **Dependabot uses a conventional-commit prefix** so its PRs satisfy the PR title linter.


jpbrodrick89 (approved):
Approved just need push production.uv.lock

PR #248, opened 2026-08-30
feat: support non-array output leaves in apply_tesseract

hozaifa1:
## What changed since the second version

@dionhaefner spelled out the expected JIT behaviour in
[this comment](https://github.com/pasteurlabs/tesseract-jax/pull/248#issuecomment-5507410858).
Points 1 and 2 already held. Point 3 is now implemented: `apply` checks the static leaves in its
response against the ones `abstract_eval` reported and warns on each one that differs. The value
the caller gets is unchanged.

## What changed since the first version

@dionhaefner said non-array outputs should be supported. This now supports them: the old error path
and its two tests are gone.

## What happens on `main`

A Tesseract whose `OutputSchema` carries a `str` or `bool` alongside its arrays cannot be used with
`apply_tesseract` at all:

```
TypeError: string indices must be integers, not 'str'
```

`abstract_eval` returns a pytree that mixes shape/dtype dicts with the plain values of the non-array
fields. The loop over the flattened avals already tolerates those leaves, and the comprehension
directly below it subscripts every leaf unconditionally:

```python
is_aval = lambda x: isinstance(x, dict) and "dtype" in x and "shape" in x
flat_avals, output_pytreedef = jax.tree.flatten(avals, is_leaf=is_aval)
for aval in flat_avals:
    if not is_aval(aval):
        continue            # <- non-aval leaves are tolerated here ...
    _check_dtype(aval["dtype"])

flat_avals = tuple(
    jax.ShapeDtypeStruct(shape=tuple(aval["shape"]), dtype=aval["dtype"])
    for aval in flat_avals   # ... and unconditionally subscripted here
)
```

## The approach

A JAX primitive can only return arrays, so a static leaf never enters the bind. `apply_tesseract`
partitions the flattened avals into arrays and statics, binds with the arrays alone, and puts the
statics back into the output pytree afterwards with `combine_args`. Two new primitive parameters
carry the partition: `static_output_mask` and `static_output_values`.

Three things follow from that, and each is where the bookkeeping could have gone wrong:

**`output_pytreedef` stays the full tree and `output_avals` holds arrays only.** Every place that
built a path-to-position map by unflattening `range(len(output_avals))` now goes through
`dummy_output_tree`, which puts `None` at each static slot. `None` is an empty pytree node, so a
static leaf disappears from anything that flattens that tree. `_pytree_to_tesseract_flat` therefore
never sees one, and every map it builds still lines up with `output_avals`. That is what keeps the
differentiable paths correct without special-casing them.

**`Jaxeract.apply` drops the static leaves from the response** before handing it back as the
primitive's outputs, and `vector_jacobian_product` fills their cotangent slots with `None` for the
same reason.

**A static output leaf carries the value `abstract_eval` reported.** Under `jit` no other value
exists, since the leaf is read at trace time and put back after the bind. A leaf whose value
depends on the input values is classified as an array.

`apply` can still return its own value for one, and it arrives too late to be used. Rather than
drop it in silence, `Jaxeract.apply` compares the static leaves in the response against the traced
values and warns on each one that differs, naming the leaf and both values:

```
UserWarning: Tesseract returned the static output ['backend'] as 'fallback' from apply, but
abstract_eval reported 'reference'. Static outputs are read at trace time, so the value from
abstract_eval is the one apply_tesseract returns and the value from apply is ignored.
```

The traced values reach `Jaxeract.apply` as a new keyword argument, passed only for `apply`, since
that is the only endpoint whose response carries static leaves at all.

## The one thing that still cannot work, and why it is JAX and not this

A `jit`-ed function still cannot *return* a static leaf:

```
TypeError: function f traced for jit returned a value of type <class 'str'>
at output component ['backend'], which is not a valid JAX type
```

That is JAX's rule about what a traced function may return, and nothing here can change it. Inside
the trace the leaf is an ordinary Python object and can be branched on, which is the useful half.
`test_a_static_leaf_is_a_python_value_inside_a_trace` pins that and says so.

## Motivating case

<https://github.com/hozaifa1/differentiable-silicon> wants `backend` and `content_hash` on every
oracle output: they are the evidence that a forward value came from the solver it is attributed to,
across a boundary where one oracle is a closed-source commercial binary and the other is
Apache-2.0. Both had to move out of the schema and into a side-channel log. That turned out to be a
reasonable audit trail. It was not the plan going in: the schema left no other place for them.

## Tests

`tests/nonarray_output_tesseract/` is unchanged: an `OutputSchema` with `y` (array), `backend`
(str) and `converged` (bool), which is the shape a provenance or status field takes in practice.
`tests/test_nonarray_outputs.py` replaces the two error-message tests:

- the static leaves come back beside the arrays, with the right values
- a static leaf is a Python value inside a `jit` trace and can be branched on
- the opaque subscript error is gone
- `jax.grad` gives `8x` for `sum(y**2)` with `y = 2x`, with and without `jit`. The statics sit
  either side of `y` in the schema, so a bookkeeping error would show up here silently, as a
  shifted or missing cotangent
- `jax.jvp` is unaffected
- `jax.vmap` is unaffected

`tests/drifting_static_tesseract/` is new: its `apply` reports `backend="fallback"` when the input
sums to a negative number, while `abstract_eval` always reports `"reference"`, so the same
Tesseract warns on one input and stays quiet on another. Four more tests cover the quiet case, the
warning and its contents, the warning under `jit`, and that the gradient is unaffected either way.

Non-array outputs are now documented in the differentiability guide, including where the value
comes from and the JAX rule that a jitted function cannot return one.

All eleven pass, and so do the other 209 tests that run on this machine. The `test_endtoend.py`
vmap tests hang locally on Windows, but they do that on an unmodified `main` too. `ruff`,
`ruff-format` and `prettier` are clean on the touched files.


hozaifa1:
@PasteurBot I have read the CLA Document and I hereby sign the CLA

hozaifa1:
recheck

dionhaefner:
Oh wow that's a pretty serious bug, thanks @hozaifa1. Non-array outputs should definitely be supported. We should get this fixed properly, I'll have a look soon unless someone else beats me to it (contributions welcome @hozaifa1, but no pressure).

hozaifa1:
Took you up on the offer. The PR now carries non-array output leaves through rather than raising a better error. I updated the title and description to match.

Short version of the approach: a JAX primitive can only return arrays, so the static leaves never enter the bind. apply_tesseract splits the flattened avals into arrays and statics, binds with the arrays alone, and puts the statics back after. Two new primitive parameters carry the split. The differentiable paths did not need special-casing: a new dummy_output_tree helper puts None at each static slot. None is an empty pytree node, so _pytree_to_tesseract_flat never sees a static, and every path-to-position map it builds still lines up with output_avals.

One limit is on JAX's side: a jit-ed function still cannot return a static leaf, because there is no JAX type for a str output. Inside the trace it is a plain Python object and can be branched on. There is a test for that.

153 tests pass locally. The 65 that error are the Docker-backed serve fixtures, which do not run on my Windows machine. CI will cover them. Open to changing any of it.

dionhaefner:
Thanks @hozaifa1! This is looking much better already.

> One limit is on JAX's side: a jit-ed function still cannot return a static leaf, because there is no JAX type for a str output. Inside the trace it is a plain Python object and can be branched on. There is a test for that.

I'd say the expected behavior is this (for JIT):

1. `abstract_eval` returns array shapes/dtypes and static leaves at trace time.
2. `apply` returns concrete values for arrays at runtime.
3. `apply` may also return static leaves at runtime, which we should check against what `abstract_eval` returned and raise a warning if they differ (will be silently ignored, static leaves are only used at trace-time).

Does that sound doable?

hozaifa1:
Yes, that is doable, and it is now implemented. Points 1 and 2 already held. Point 3 is the new part.

`apply` now compares the static leaves in its response against the ones `abstract_eval` reported at trace time, and warns on each one that differs:

```
UserWarning: Tesseract returned the static output ['backend'] as 'fallback' from apply,
but abstract_eval reported 'reference'. Static outputs are read at trace time, so the
value from abstract_eval is the one apply_tesseract returns and the value from apply
is ignored.
```

The warning names the leaf and both values, so the author can see which field drifted. The caller still gets the value from `abstract_eval`. The runtime one from `apply` is ignored, exactly as you described.

`Jaxeract.apply` takes the traced values as a new keyword argument. It's passed only for `apply`, since that's the only endpoint whose response carries static leaves at all. The comparison sits next to the code that was already dropping them.

Added a test tesseract whose `apply` returns a different backend depending on the input sign, so one input warns and another does not. Tests cover the quiet case, the warning, the warning under `jit`, and that gradients are unaffected. Also documented in the differentiability guide, alongside the JAX rule that a jitted function cannot return a static leaf.

220 tests pass locally. The `test_endtoend.py` vmap tests hang on my Windows machine, but they do that on an unmodified `main` too, so CI will have to be the judge there. The workflow runs are sitting at `action_required` if you can approve them.

docs/content/handling-differentiability.md
```diff
@@ -133,3 +133,52 @@ Note that the cotangent/tangent pytree structure must always match the function'
...
+
+**A jitted function cannot return one.** This is JAX's own rule about what a traced
+function may return — there is no JAX type for a `str` output:
+
+```
+TypeError: function f traced for jit returned a value of type <class 'str'>
+at output component ['backend'], which is not a valid JAX type
+```
+
+Consume the field inside the trace, as above, and return the arrays.
```
```suggestion
```

Superfluous imo

docs/content/handling-differentiability.md
```diff
@@ -133,3 +133,52 @@ Note that the cotangent/tangent pytree structure must always match the function'
...
+
+```python
+class OutputSchema(BaseModel):
+    y: Differentiable[Array[(3,), Float64]]
+    backend: str = "reference"
+    converged: bool = True
+```
+
+`apply_tesseract` returns these alongside the arrays, and inside a `jit` trace they
+are ordinary Python objects, so you can branch on them:
```
dionhaefner: Should spell out that this is jit-only, in fact in eager mode `abstract_eval` is never called

docs/content/handling-differentiability.md
```diff
@@ -133,3 +133,52 @@ Note that the cotangent/tangent pytree structure must always match the function'
...
+
+```python
+@jax.jit
+def f(x):
+    out = apply_tesseract(tess, {"x": x})
+    scale = 1.0 if out["converged"] else 0.0   # a trace-time branch
+    return out["y"] * scale
+```
+
+Two consequences are worth knowing about.
```
dionhaefner: Ready very LLM-y, can we deslop this text a bit?

tesseract_jax/primitive.py
```diff
@@ -1074,15 +1122,38 @@ def apply_tesseract(
...
+        # An OutputSchema may carry non-array fields alongside its arrays: a
+        # backend name, a convergence flag, a content hash. A JAX primitive can
+        # only return arrays, so those leaves never enter the bind. They are
+        # read from abstract_eval, held aside as static primitive parameters,
+        # and put back into the output pytree once the bind has returned.
+        #
+        # One consequence is worth stating plainly: a static output leaf is the
+        # value abstract_eval reported, not the value apply returned. Under jit
+        # no other value exists, and a leaf whose value depends on the input
+        # values is an array, not a static.
```
dionhaefner:
```suggestion
```

tesseract_jax/tree_util.py
```diff
@@ -152,3 +153,64 @@ def _pytree_to_tesseract_flat(
...
+    """Whether two static leaves disagree.
+
+    Static leaves are decoded response data, so ``!=`` settles it. The guard is
+    for a leaf whose comparison does not answer with a bool -- an array-valued
+    one, say -- where identity is the only question left that still has an answer.
+    """
+    try:
+        return bool(returned != expected)
+    except (TypeError, ValueError):
+        return returned is not expected
```
dionhaefner, line 206: This could use some qualifying regarding when it can occur. Served Tesseracts can only return JSON-native types so should be unreachable. Tesseract using `from_tesseract_api` *could* technically return something else in some edge cases, although it's not clear to me under which circumstances.

tesseract_jax/tree_util.py
```diff
@@ -152,3 +153,64 @@ def _pytree_to_tesseract_flat(
...
+    be used: ``apply_tesseract`` hands back the value ``abstract_eval`` reported.
+    A Tesseract that returns a different one from ``apply`` is therefore doing
+    something the caller cannot observe, and silence would hide that.
+    """
+    for path, returned, expected in zip(
+        paths, returned_values, expected_values, strict=True
+    ):
+        if not _leaves_differ(returned, expected):
+            continue
+        warnings.warn(
```
dionhaefner, line 226: I wonder how this looks at the other end, given we're raising this warning from an XLA callback. Is it rendered as expected?

dionhaefner (approved):
Very nice, thanks @hozaifa1 ! Just a few comments from my end.

jpbrodrick89:
Nice work @hozaifa1, thanks for catching this important limitation! Just a couple things from my side:

1. I think we should add a line on this in the README when discussing sharp edges and abstract_eval: something like "Note that, under jit any non-array (i.e. static) outputs will be fully determined by `abstract_eval`." 
2. I'm a bit concerned about the cost of the static value comparisons, I know we can silence `UserWarning`'s but that doesn't stop the work of actually doing the comparisons. @dionhaefner what's your take here, should we allow an opt out either as a kwarg in `apply_tesseract` (something else we need to pass around), or a jax-like config variable that could either be set as an environment variable or just after module import?

hozaifa1:
Both done.

On the README: there is now a "Non-array outputs come from `abstract_eval`" bullet under Sharp edges. It says the value is fixed at trace time, that whatever `apply` returns for such a field arrives too late to be used, and that a field whose value depends on the input values belongs in the schema as an array.

On the cost: there is a config object now, `tesseract_jax.config`, with one setting.

```python
import tesseract_jax

tesseract_jax.config.check_static_outputs = False        # rest of the process

with tesseract_jax.config.set(check_static_outputs=False):   # or one block
    out = apply_tesseract(tess, {"x": x})
```

`TESSERACT_JAX_CHECK_STATIC_OUTPUTS=0` sets the default at import.

Turning it off skips the work, not only the warning. The comparison itself is cheap. The cost is that naming a field in the warning needs `tree_flatten_with_path` on the response, where plain `tree_flatten` would do. With the check off, `Jaxeract.apply` takes the plain path, and the dispatch functions stop unpacking `static_output_values` per call. A Tesseract with no static outputs at all was already paying nothing, since the whole block is behind `any(static_output_mask)`.

I went with the config variable over a kwarg on `apply_tesseract` because of the plumbing you mentioned: it would have to be threaded through the primitive params, and it is the sort of thing you set once for a program rather than per call. Happy to add the kwarg as well if you would rather have both.

New `tests/test_config.py` covers the env var parsing, the defaults, assignment, the block form and its restore-on-failure, and the two unknown-name errors. Four more tests in `test_nonarray_outputs.py` cover the check being off end to end: silence on a drifting Tesseract, the gradient unaffected, the setting restored after the block, and the unknown-setting error. Docs updated in the differentiability guide next to the warning.

dionhaefner:
Thanks @hozaifa1. I'm not too hot on a config here, it feels like per-request is more meaningful for this feature – consider for example a pipeline where there's a single "naughty" Tesseract that returns non-deterministic outputs as a side effect, or the case where a comparison is known to be expensive. In those cases I'd want to disable the check selectively, not globally.

Since we have the `DispatchParams` object now it should also be a lot less added code than an entire config module.

Happy to merge once that + my review comments above are addressed.

jpbrodrick89:
@dionhaefner lets discuss tomorrow, I'm having second thoughts on this. Personally, I'd lean towards keeping some kind of global config to allow easy switching between debug and production mode. I do buy your argument and think it warrants more discussion.

@hozaifa1 one of us will get back to you on the preferred design tomorrow after we've discussed. Apologies for the delay.

jpbrodrick89:
Hi @hozaifa1 after discussion with Dion we've decided to go with env vars (`TESSERACT_JAX_CHECK_STATIC_OUTPUTS` as you had it) for global config _and_ kwargs for per-Tesseract config. We feel that an actual config object and config context managers is not necessary at this stage. How does that sound?

hozaifa1:
Env var plus kwarg, done. The config module and tests/test_config.py are removed.

apply_tesseract takes check_static_outputs now:

```python
out = apply_tesseract(tess, {"x": x}, check_static_outputs=False)
```

It rides on DispatchParams, so the plumbing is one field. Left at None it reads TESSERACT_JAX_CHECK_STATIC_OUTPUTS, and the kwarg wins when both are set. The variable is read on each call rather than once at import, so a program can change it at runtime.

On the review comments:


Docs: the "a jitted function cannot return one" block is gone, and the rest is trimmed. The second paragraph of the comment in primitive.py is gone too.

_leaves_differ: the docstring now says a served Tesseract cannot reach the identity fallback, because JSON has no type whose != returns a non-bool. from_tesseract_api can reach it, since it hands back whatever the Python function built. An abstract_eval that reports a numpy array for a field makes that field static, and a != b on two arrays is an array, which bool() refuses. There is a unit test on it now.

Eager mode: I do not think the jit-only framing holds here. apply_tesseract calls abstract_eval whenever the endpoint exists, with or without jit, so the static leaf comes from abstract_eval in both cases and the drift warning fires eagerly too. test_a_static_leaf_that_apply_disagrees_with_is_warned_about is the eager case. I documented it that way. If you would rather eager mode passed apply's value straight through, say so and I will change it, though then the same call returns different values with and without jit.

Warning rendering from the callback: it renders, and identically in both modes. The location it names is a library line, since under jit there is no user frame on the stack at all:

```
tesseract_jax/tesseract_compat.py:150: UserWarning: Tesseract returned the static output ['backend'] as 'fallback' from apply, but abstract_eval reported 'reference'. ...
```

So the field name in the message is the part that points at the problem. I can drop the stacklevel if you would rather it named warn_on_static_output_drift itself.

dionhaefner:
Pushed some tweaks so `abstract_eval` is now never called in eager mode and always called (without fallback) during tracing. I realize now this is a behavior change, and it's relying on a private JAX symbol, but I kind of like the clean split of eager vs. traced. Thoughts @hozaifa1 @jpbrodrick89 ?

hozaifa1:
I like the split. The private import earns its place. I checked the three cases against the previous commit.

Eager now returns `apply`'s own static value and does not warn, which is the right answer. Under `jit`, `grad` and `vmap` the value still comes from `abstract_eval`; the drift warning fires as before.

The case that justifies `trace_state_clean` over a tracer check is a concrete array closed over by the traced function. Your import puts that case on the traced path. The tracer check alone calls it eager, so the call dies in `tesseract_dispatch_abstract_eval` on `for aval in params.output_avals`, which is None on that path. That means the `ImportError` fallback crashes there. `pragma: no cover` means nothing would catch it if the symbol moves. I would rather let the `ImportError` propagate at import time, or pin a jax floor, than keep a fallback that turns a moved symbol into a `TypeError` inside a trace.

I went looking for a public equivalent. `jax.core` does not export `trace_state_clean`. `jax.extend.core` only has `TraceTag`, `find_top_trace`, `get_opaque_trace_state`, `set_current_trace` and `take_current_trace`, none of which answer the question. So the private import is the only option today; the comment could say so.

`_check_dtype` now only runs on the traced path. With `jax_enable_x64` off, an eager call on a Tesseract with a float64 output used to raise "Cannot return 64-bit values when jax_enable_x64 is disabled"; it now returns a float64 numpy array quietly. It raised on `0256e7b` and does not on `da38b44`.

Full suite passes here.

jpbrodrick89:
I very much like the _idea_ of skipping abstract_eval but we need to think very carefully about the long-term reason why we're doing this:

1. Allow skipping an `abstract_eval` round trip: nice-to-have, especially in high latency scenarios but probably not worth a consistency break on its own
2. Eager mode as a "feature", meaning we can support Tesseracts without `abstract_eval` perhaps with dynamic shapes or dynamic non-arrays. This sounds very attractive and potentially revolutionary, but signs us up for some significant follow-up work. Right now, @dionhaefner's changes mean we ONLY support this "feature" for `apply` which is not very useful as someone in that case could just use the `tesseract-core` SDK rather than tesseract-jax. Extending this to allow `jvp` support is trivial, `vjp` is harder but I think its do-able (by extracting concrete shapes from the `apply` call when creating the `vjp_fun` closure), non-sequential `vmap` is probably hardest and we might have to skip that. If we can commit to supporting at least two function transformations (ideally more) in the near future then I think this inconsistency will more than pay for itself, if we're not sure we should consider introducing the breaking change later when we make our mind up.

Another thing to note is that skipping `abstract_eval` in eager mode will make using `cuda_ipc` in eager mode with XLA a bit trickier but a pure Python path will still be available. The current `cuda_ipc` PR will be completely unaffected by this because it only adds support through a lowering which only fires under jit. In the future if we are sold on the "eager as feature" idea, eager with `cuda_ipc` will probably be highly desirable and the most obvious route for that is probably using `with client.with_cuda_ipc()` and calling in pure Python (skip the cuda shim) directly and convert the returned `IPCDeviceArray`s with `from_dlpack`. The challenge would be inputs (jax.Array exposes `__dlpack__` but not `__cuda_array_interface__`) but we can probably work a way around that.

A few bits of feedback:

1. A deleted test means we no longer test that the warning fires at every callback time rather than trace time.
2. No tests are asserting that `abstract_eval` isn't called eagerly with a spy.
3. `trace_state_clean is None` fallback not covered in tests.
4. has_func_transformation and is_eager use different definitions of "in a transformation"

dionhaefner:
The only motivation is not deeper than "Tesseracts should behave identically in eager mode whether they define `abstract_eval` or not". Before I took over that contract was violated, since T-JAX would always dispatch to `abstract_eval`, then take a fallback path if that dispatch failed. Yes it's kind of pointless to use T-JAX in that case but I figured you should still be able to.

> The current cuda_ipc PR will be completely unaffected by this because it only adds support through a lowering which only fires under jit.

I don't think that's true, the lowering should always fire.

> If we can commit to supporting at least two function transformations (ideally more) in the near future 

I'd say no plans to do so, and going down this route warrants a strong motivating use case.

dionhaefner:
Also is this really a breaking change? I can't come up with a case where this would be a regression (other than "I relied on my broken `abstract_eval` to raise and caught the exception, now it's suddenly passing wtf?!").

hozaifa1:
All four points are done.

1. A test jits once and calls twice, and expects the warning both times. A trace-time comparison would only fire on the first.

2. A spy on `Jaxeract.abstract_eval` shows no eager calls and one call under jit.

3 and 4 turned out to be the same bug. `has_func_transformation` scanned the inputs for tracers, `is_eager` asked `trace_state_clean`, so a traced function calling `apply_tesseract` on values it closed over passed the `abstract_eval` endpoint check and then failed with "RuntimeError: Endpoint abstract_eval not found in Tesseract API". Both now come from one `_in_transformation` call, so they cannot disagree.

The fallback is covered now. It no longer crashes. Without `trace_state_clean` the tracer scan decides, which is right for any call that passes a traced argument. The one case it cannot see used to die on `for aval in params.output_avals` with `output_avals` None, and now says what happened. The test sets `_trace_state_clean` to None and walks the eager call, the traced call and that case.

@dionhaefner on whether it is breaking: I found one user-visible change. I now think the new behaviour is the better one, so I am correcting what I said above. `_check_dtype` only runs on the traced path now, so with `jax_enable_x64` off an eager call on a Tesseract with a `float64` output used to raise "Cannot return 64-bit values when jax_enable_x64 is disabled" and now returns the array. The eager path hands back `numpy` either way, before and after, so that error was rejecting a value it was going to return as `numpy` regardless. Nothing in the docs mentions it, which is the only thing I would add.

Full suite passes here, 248 tests.

jpbrodrick89:
> I can't come up with a case where this would be a regression (other than "I relied on my broken abstract_eval to raise and caught the exception, now it's suddenly passing wtf?!").

I agree you're not really introducing a breaking change, but after this PR lands changing our mind on this does become a breaking change, here are the three options we can decide between today:

1. Prioritising consistency in eager mode between Tesseracts that do and don't have an `abstract_eval` at the price of consistency between jit and eager mode by dropping an `abstract_eval` roundtrip in eager mode for `apply`. (post Dion's changes)
2. Prioritising consistency between jit and eager mode at the price of consistency in eager mode between Tesseracts that do and don't have an `abstract_eval`. (pre Dion's changes)
3. Prioritising consistency in all three cases at the cost of dropping support (breaking change) for Tesseracts without `abstract_eval` in eager mode.

Only 3 is a breaking change today, but switching between the options after this PR lands will be a more substantial breaking change because it will change how (previously unsupported) non-array outputs are treated and not just round trip counts and whether an error is raised.

> > The current cuda_ipc PR will be completely unaffected by this because it only adds support through a lowering which only fires under jit.
> 
> I don't think that's true, the lowering should always fire.

It is true, and this is REALLY important to understand. For a lot of JAX primitives (`sin_p`, `lu_p`) you're right, but the reason is because the eager implementation is _explicitly_ made equal to the lowering because the primitve's `def_impl` is simply `dispatch.apply_primitive` which actually compiles the primitive (first evaluating abstractly) under the hood and then evaluates the compileda primitive. However, `tesseract_dispatch_p`'s `def_impl` is NOT `dispatch.apply_primitive` but a pure Python implementation. As such, the lowering is never used in eager mode: a CPU dispatch will be silently performed even when `cuda_ipc=True` and all other boxes are checked.

As such, the choice we make here today has real implication on how we could support `cuda_ipc` in eager mode.

a) If we plan on switching `def_impl` to be `dispatch.apply_primitive` so that we use the lowering automatically then ONLY option 3 above makes sense (GPU lowering with cuda shim always requires abstract eval)
b) If we implement a separate pure Python GPU path then we support any of the 3 options (and arguably Dion's 1 makes most sense).
c) We don't bother supporting eager GPU (interpretation: eager mode is not a feature). In this case I would also lean towards 3, if eager mode is not a feature or intended usage then we should just raise loudly  so users don't start sketching a workflow that they only find out later doesn't work. This also makes switching to (a) easy if we change our mind in the future.

> going down this route warrants a strong motivating use case.

The motivating cases are where writing an `abstract_eval` is not possible (shapes are dynamic or we have an existing black box third party Tesseract that doesn't have one) and we have a workflow that doesn't require `jit` (for example typical `optax` or `scipy` workflows don't but `optimistix` ones do as they're implemented in a `fori_loop` or `while` loop). For our immediate interests the most compelling case is likely a parameterised CAD model to mesh where the number of cells depends on the values of the CAD model.

<details>
<summary>
Note I have validated this works in principle for a simple area to perimeter maximisation of a rectangle by summing a dynamically shaped grid of cell areas:
</summary>

```python
import jax
import jax.numpy as jnp
import numpy as np
from jax import extend as jex
from jax.interpreters import ad

jax.config.update("jax_enable_x64", True)

grid_p = jex.core.Primitive("grid")
grid_p.multiple_results = True


def abstract_eval_endpoint(*args, **kwargs):
    """Stand-in for a Tesseract's remote abstract_eval. Never reached."""
    raise AssertionError("abstract_eval endpoint was called")


@grid_p.def_impl
def _grid_impl(*args, endpoint, out_shapes):
    if endpoint == "apply":
        w, h = float(args[0]), float(args[1])
        nx, ny = int(np.ceil(w / 0.1)), int(np.ceil(h / 0.1))  # cells depend on values
        return [np.full((nx, ny), w * h / (nx * ny))]          # per-cell areas
    if endpoint == "jvp":
        raise AssertionError("jvp endpoint executed -- transposition should consume it")
    if endpoint == "vjp":
        w, h, ct = float(args[0]), float(args[1]), np.asarray(args[2])
        per_cell = ct.sum() / ct.size  # the cotangent carries the cell count
        return [np.float64(per_cell * h), np.float64(per_cell * w)]
    raise ValueError(endpoint)

@grid_p.def_abstract_eval
def _grid_abstract_eval(*avals, endpoint, out_shapes):
    """Local: reads shapes the JVP rule already learned from a concrete apply."""
    if out_shapes is None:
        # This path should only be taken under jit
        out_shapes = abstract_eval_endpoint(*avals, endpoint=endpoint)
    return [jax.core.ShapedArray(shape, np.dtype("float64")) for shape in out_shapes]


def _grid_jvp(primals, tangents, *, endpoint, out_shapes):
    """Primals are concrete in eager mode, so the true output shape is known here."""
    w, h = primals
    assert not isinstance(w, jax.core.Tracer), "primal must be concrete"
    (primal_out,) = _grid_impl(w, h, endpoint="apply", out_shapes=None)
    (tangent_out,) = grid_p.bind(
        w, h, *tangents, endpoint="jvp", out_shapes=(primal_out.shape,)
    )
    return [primal_out], [tangent_out]


def _grid_transpose(cotangents, w, h, dw, dh, *, endpoint, out_shapes):
    """Rewrites the staged jvp equation into a vjp call."""
    (ct,) = cotangents
    ct_w, ct_h = grid_p.bind(w, h, ct, endpoint="vjp", out_shapes=((), ()))
    return [None, None, ct_w, ct_h]  # the primals are not linear positions

ad.primitive_jvps[grid_p] = _grid_jvp
ad.primitive_transposes[grid_p] = _grid_transpose


def grid(w, h):
    (areas,) = grid_p.bind(w, h, endpoint="apply", out_shapes=None)
    return areas


def loss(u):
    """Negative area of a rectangle of fixed perimeter 4, summed over cells."""
    w = 1.0 + jnp.tanh(u)        # w in (0, 2)
    h = 2.0 - w                  # so w + h == 2, perimeter == 4
    return -jnp.sum(grid(w, h))  # == -w*h, however many cells there are


u = jnp.array(-1.1)
print(f"{'step':>4} {'w':>7} {'h':>7} {'cells':>7} {'area':>9} {'grad':>10}")
for step in range(8):
    value, gradient = jax.value_and_grad(loss)(u)
    w = 1.0 + jnp.tanh(u)
    nx, ny = grid(w, 2.0 - w).shape
    print(
        f"{step:>4} {float(w):>7.4f} {float(2.0 - w):>7.4f} {f'{nx}x{ny}':>7} "
        f"{-float(value):>9.5f} {float(gradient):>10.5f}"
    )
    u = u - 0.6 * gradient

w = float(1.0 + jnp.tanh(u))
print(f"\nconverged: w={w:.5f} h={2 - w:.5f} (square = 1.0), area={-float(loss(u)):.6f} (max = 1.0)")

finite_diff = (loss(u + 1e-6) - loss(u - 1e-6)) / 2e-6
print(f"gradient check: autodiff {float(jax.grad(loss)(u)):+.4e} vs finite-diff {float(finite_diff):+.4e}")

for label, fn in [("jit(loss)      ", lambda: jax.jit(loss)(u)),
                  ("jit(grad(loss))", lambda: jax.jit(jax.grad(loss))(u))]:
    try:
        fn()
    except Exception as exc:
        print(f"{label}: {type(exc).__name__} - {str(exc)[:60]}")
```

Output (jax 0.10.0, CPU):

step       w       h   cells      area       grad
   0  0.1995  1.8005    2x19   0.35920   -0.57508
   1  0.3619  1.6381    4x17   0.59284   -0.75657
   2  0.7078  1.2922    8x13   0.91460   -0.53456
   3  1.0197  0.9803   11x10   0.99961    0.03943
   4  0.9961  1.0039   10x11   0.99998   -0.00787
   5  1.0008  0.9992   11x10   1.00000    0.00157
   6  0.9998  1.0002   10x11   1.00000   -0.00031
   7  1.0000  1.0000   11x10   1.00000    0.00006

converged: w=0.99999 h=1.00001 (square = 1.0), area=1.000000 (max = 1.0)
gradient check: autodiff -1.2583e-05 vs finite-diff -1.2584e-05
jit(loss)      : AssertionError - abstract_eval endpoint was called
jit(grad(loss)): AssertionError - primal must be concrete
</details>

Personally, I'm not sure how _strong_ this case is given the (manageable but not neglgible) engineering involved but it seems a way to get around a lot of the inherent limitations in JAX in an actually useful way.

@hozaifa1 sorry for the runaround, but it turns out the decision between 1, 2 and 3 has real implications once this PR lands.

dionhaefner:
Ah I thought we already had `dispatch.apply_primitive` in here (as e.g. mpi4jax does).

Thanks for the useful analysis. My gut feeling is we should adopt apply_primitive and do (3) now, then potentially soften it up later if there's a concrete application that needs it. To me JAX-without-jit isn't really JAX, and as you said yourself people can just use the T-Core client.

For this PR this would mean we could merge as-is then drop support for Tesseracts without abstract_eval in a follow-up?

jpbrodrick89:
Agreed, as long as don't do a release between this PR and doing 3.

dionhaefner:
sgtm, and thanks @hozaifa1 !

PR #252, opened 2026-09-03
fix: keep dotted dict keys whole when matching schema templates

### LINKED ISSUE #251: Dotted dict keys are misclassified as non-differentiable in _merge_path
## Description

A `dict[str, Differentiable[...]]` field whose keys contain dots — the natural shape for a state dict, e.g. `encoder.layer.0.weight` — is reported differentiable and applies fine, but a gradient through it either fails with a message that points nowhere near the cause or is silently classified as non-differentiable.

The dict-key gradient path is spelled `params.{<key>}` on the wire. `_merge_path` in [`tesseract_jax/tree_util.py`](https://github.com/pasteurlabs/tesseract-jax/blob/main/tesseract_jax/tree_util.py) splits **both** the concrete path and the template on every `.` before comparing them:

```python
explicit_parts = explicit_path.split(".")
...
template_parts = array_path.split(".")
if len(template_parts) != len(explicit_parts):
    continue
```

A concrete path built from a dotted key has more parts than its single-`{}` template, so the length check never matches and `_merge_path` returns `(explicit_path, None)`. The `None` matched-template is what gets read downstream as "this field is not differentiable", which is why the surfaced error talks about a tangent rather than the key.

```python
_merge_path('params.plain',           ['params.{}'])  ->  ('params.{plain}', 'params.{}')  # ok
_merge_path('params.a/b',             ['params.{}'])  ->  ('params.{a/b}',   'params.{}')  # ok (slashes are fine)
_merge_path('params.layer.0.weight',  ['params.{}'])  ->  ('params.layer.0.weight', None)  # dotted key: no match
```

This is the tesseract-jax side of pasteurlabs/tesseract-core#705. The tesseract-core half — widening the schema pattern to `\{[^.]+\}` so slashed/colon'd keys resolve — landed in pasteurlabs/tesseract-core#707. Dotted keys were deliberately left refused there because they need a brace-aware split, which is the part that lives here.

The fix is to make the split in `_merge_path` brace-aware, so a `{...}` segment is treated as one part regardless of the dots inside it, rather than splitting on every `.`.

## Steps to reproduce

```python
import jax
import numpy as np
from tesseract_core import Tesseract
from tesseract_jax import apply_tesseract

# A Tesseract with an input field:
#   params: dict[str, Differentiable[Array[(3,), Float32]]]

for key in ("plain", "a/b", "layer.0.weight"):
    inputs = {"params": {key: np.ones(3, dtype=np.float32)}}
    try:
        g = jax.grad(lambda x: apply_tesseract(t, x)["result"].sum())(inputs)
        print(f"{key!r:20} OK")
    except Exception as e:
        print(f"{key!r:20} {type(e).__name__}: {e}")
```

## Logs

```bash
'plain'              OK
'a/b'                jsonschema error: String should match pattern   # fixed by tesseract-core#707
'layer.0.weight'     Non-symbolic-zero tangent provided for non-differentiable input 'params.layer.0.weight'
```

(The `a/b` line now resolves once the tesseract-core#707 pattern fix is picked up; the dotted line is the remaining tesseract-jax bug.)

## OS

Linux, Mac

## Tesseract + Tesseract-JAX version

tesseract-core >= 1.12.0 (with pasteurlabs/tesseract-core#707), tesseract-jax latest `main`

singhharsh1708:
A dict key may contain dots, so joining the pytree segments into a path and splitting it again cannot tell `{"a": {"b.c": v}}` from `{"a": {"b": {"c": v}}}`. That is why a dotted key missed its template and came back as a non-differentiable input.

`_pytree_to_tesseract_flat` now passes `_merge_path` the segments it already had rather than a joined string. `_merge_path` still takes a string, splitting on the dots outside braces, and no longer braces a segment that is already braced, which is what `batching` needs when it re-merges a path this function produced.

```
jax.grad through params.{layer.0.weight}
  before  ValueError: Non-symbolic-zero tangent provided for non-differentiable input
  after   grad=[2. 2. 2.]
```

Needs the pattern from pasteurlabs/tesseract-core#707 to reach the runtime, so I verified against tesseract-core `main` rather than the released 1.12.0. 283 passed.

Fixes #251.

### DIFF
--- a/tesseract_jax/tree_util.py
+++ b/tesseract_jax/tree_util.py
@@ -72,20 +72,53 @@ def unflatten_args(
     return result
 
 
+def _split_path(path: str) -> list[str]:
+    """Split a path on the dots that separate segments.
+
+    A dot inside ``{...}`` belongs to the key, so ``a.{b.c}`` is two segments
+    rather than three.
+    """
+    parts: list[str] = []
+    buf: list[str] = []
+    depth = 0
+    for char in path:
+        if char == "{":
+            depth += 1
+        elif char == "}":
+            depth -= 1
+        if char == "." and depth == 0:
+            parts.append("".join(buf))
+            buf = []
+        else:
+            buf.append(char)
+    parts.append("".join(buf))
+    return parts
+
+
 def _merge_path(
-    explicit_path: str, array_paths: Iterable[str]
+    explicit_path: str | Sequence[str], array_paths: Iterable[str]
 ) -> tuple[str, str | None]:
     """Merges and formats explicit path with array paths containing templates.
 
     Returns a tuple of (formatted_path, matched_template) where matched_template
     is the template string that matched, or None if no template matched.
 
+    ``explicit_path`` may be given as the already-joined string or as the
+    segments it was built from. A dict key is free to contain dots, so passing
+    the segments is the only way to say where one ends; joining first and
+    splitting again cannot tell ``{"a": {"b.c": v}}`` from ``{"a": {"b": {"c": v}}}``.
+
     Examples:
         _merge_path('alpha.beta.x', ['alpha.beta.{}']) -> ('alpha.beta.{x}', 'alpha.beta.{}')
         _merge_path('delta.[2]', ['delta.[]']) -> ('delta.[2]', 'delta.[]')
         _merge_path('epsilon.k', ['alpha.{}']) -> ('epsilon.k', None)
+        _merge_path(['params', 'a.b'], ['params.{}']) -> ('params.{a.b}', 'params.{}')
     """
-    explicit_parts = explicit_path.split(".")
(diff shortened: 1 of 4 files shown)


dionhaefner:
Thanks @singhharsh1708 ! Let's let this sit until the Tesseract-Core fix is released, then add one end-to-end test?

singhharsh1708:
Added it in 49f01ba, skipped until the core pattern is importable so it turns itself on when that releases. It fails on main with the non-differentiable-input error from the issue.

singhharsh1708:
Carried over the version-keyed skip you asked for on tesseract-torch#50, since this had the same import probe. It now compares tesseract_core.__version__ against 1.12.0 and the skip reason names the version in use.

dionhaefner (approved):
Thanks @singhharsh1708

PR #253, opened 2026-09-03
feat: select the on-device transport by name

dionhaefner:
#### Relevant issue or PR

Follows #239 (GPU cuda_ipc device-to-device interop) and is based on its branch. Pairs with the device-transport interface in tesseract-core (pasteurlabs/tesseract-core#722).

#### Description of changes

#239 added a single on-device GPU transport, opted into with `apply_tesseract(..., cuda_ipc=True)`. tesseract-core now exposes the transport machinery behind a named, pluggable interface, so this generalizes the opt-in to select a transport by name:

```python
apply_tesseract(tess, inputs, device_transport="cuda_ipc")
```

`Jaxeract` stores the transport name instead of a bare boolean:
- its equality/hash key on the name, so two calls that use *different* transports don't get commoned up by XLA into one custom call;
- its scoped context manager (renamed `cuda_ipc()` → `device_transport_encoding()`, old name kept as an alias) sets the client's `_output_format` and `Accept` header to `application/json+<name>` for the duration of one dispatch.

`cuda_ipc=True` is kept as the exact back-compatible spelling of `device_transport="cuda_ipc"`, so existing callers are unaffected.

An unsupported transport name is rejected in the constructor. This matters because the GPU (FFI) lowering is currently cuda_ipc-specific: without the check, a name like `device_transport="foo"` would set the on-device flag, route into that lowering, and send an `Accept` the server has no backend for. The set of accepted names lives in one place (`_SUPPORTED_TRANSPORTS`) and grows as the FFI path learns to drive further transports.

#### Testing done

- New `tests/test_transport_selection.py` (GPU-free): transport-name selection, the `cuda_ipc=True` alias, default host round-trip, unsupported-name rejection, and that equality/hash key on the transport. 6 pass.
- `tests/test_gpu_direct.py` extended with an end-to-end `device_transport="cuda_ipc"` case asserting it stays on-device and matches the `cuda_ipc=True` path bit-for-bit; the existing `-m gpu` suite (apply/grad/jvp/vjp/jacobian) still passes on a GPU box. These run on the `tests-gpu` CI job.
- `pre-commit` clean.


dionhaefner:
Now folded into #239

PR #256, opened 2026-09-07
refactor: bundle primitive dispatch kwargs into a DispatchParams dataclass

dionhaefner:
#### Description of changes

The `tesseract_dispatch` primitive threaded ~13 descriptors — pytreedefs, static/tangent masks, output avals, the `Jaxeract` client, `eval_func`, and the vmap/jacobian options — as individual keyword arguments through every primitive rule (abstract-eval, jvp, transpose, impl, lowering, batching), every batching strategy, and every `Jaxeract` endpoint method. The signatures had grown to a dozen-plus parameters each and every `bind()` call restated the whole set, which made them noisy to read and easy to get out of sync.

This bundles them into a single frozen `DispatchParams` dataclass (new `dispatch_params.py`) carried as one bind parameter. The dataclass also absorbs two bits of repetition: an `n_primals` property (previously recomputed inline in ~8 places) and a `replace()` helper for the derived binds. No behavior change — purely a signature refactor.

The dataclass is frozen with all-hashable fields, which is what keeps JAX's bind-param equality intact and preserves XLA's CSE folding of identical Tesseract calls (`Jaxeract` still supplies its own `__eq__`/`__hash__`).

#### Testing done

- Full suite: `279 passed`.
- Ruff check + format clean.
- Confirmed the CSE invariant still holds via the existing `test_identical_calls_are_commoned_up` / `test_distinct_calls_are_not_commoned_up` tests.

### DIFF
--- a/tesseract_jax/dispatch_params.py
+++ b/tesseract_jax/dispatch_params.py
@@ -1,0 +1,74 @@
+# Copyright 2025 Pasteur Labs. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""The parameter bundle threaded through the ``tesseract_dispatch`` primitive.
+
+Every rule of the primitive (abstract eval, jvp, transpose, batching, lowering,
+impl) and every batching strategy needs the same set of descriptors: how the
+flat operands map back to the input pytree, what the outputs look like, which
+endpoint to call, and how to handle vmap. Passing them as a dozen individual
+keyword arguments made the signatures unwieldy, so they live together in a
+single frozen dataclass that travels as one primitive bind parameter.
+
+The dataclass is frozen and every field is hashable, which is what lets JAX
+compare two binds for equality and lets XLA common up identical Tesseract
+calls (see the CSE note in ``primitive.py``).
+"""
+
+from dataclasses import dataclass, replace
+from typing import TYPE_CHECKING, Any, Literal
+
+from jax import ShapeDtypeStruct
+from jax.tree_util import PyTreeDef
+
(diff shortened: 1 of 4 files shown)


tesseract_jax/dispatch_params.py
```diff
@@ -0,0 +1,74 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""The parameter bundle threaded through the ``tesseract_dispatch`` primitive.
+
+Every rule of the primitive (abstract eval, jvp, transpose, batching, lowering,
+impl) and every batching strategy needs the same set of descriptors: how the
+flat operands map back to the input pytree, what the outputs look like, which
+endpoint to call, and how to handle vmap. Passing them as a dozen individual
+keyword arguments made the signatures unwieldy, so they live together in a
+single frozen dataclass that travels as one primitive bind parameter.
```
jpbrodrick89: AI smell/overly convoluted.

```suggestion
Every rule of the primitive and every batching strategy needs the same 
set of descriptors. Passing them as a single frozen dataclass is neater than
a dozen individual keyword arguments on every rule signatures.
```

jpbrodrick89 (approved):
This is a nice, convenient and mnimally controversial improvement. There may be more elegant designs, e.g. using [HiJax](https://docs.jax.dev/en/latest/hijax_types.html) (which is experimental and I believe not backwards compatible) or making the primitive itself a callable class/equinox Module, but getting that right takes a lot more thought and I'm not sure it's worth the brainspace. Let's go with this for a nice quick win! Thanks @dionhaefner

Issue #258, opened 2026-09-07
Spurious RuntimeWarning (or opaque CpuCallback error) for gradients through a Tesseract with an integer array input


jpbrodrick89:
### Description

Taking a gradient through a `jit`-ed function whose Tesseract has a **non-differentiable integer array input** emits a spurious `RuntimeWarning: invalid value encountered in cast`. Gradients are correct; JAX returns the expected zero-byte `float0` cotangent for the integer input, so the warning has no effect on results.

Under warnings-as-errors — including this project's own `filterwarnings = ["error"]` in `pyproject.toml`, or any user running `-W error` — it is not a warning but a failure, surfacing as an opaque `jax.errors.JaxRuntimeError: INTERNAL: CpuCallback error calling callback` with the cause buried at the bottom of the traceback.

**Cause:** the discarded gradient slot for a non-differentiable input is filled with `np.full(shape, np.nan, dtype=dtype)` (`tesseract_compat.py:313` on `main`). For an integer dtype that cast is invalid: NumPy warns and yields `0`. The value is never read — the slot's tangent was a symbolic zero, so it is never one of the `UndefinedPrimal` positions JAX transposes — so only the warning is observable.

**Trigger requires `jit`.** Eagerly, JAX passes an integer argument through as a concrete value (its tangent space is `float0`), so `is_static_mask` (`primitive.py:1203`, which tests tracer-ness, not differentiability) routes it to `static_args` and the placeholder never runs. Under `jit` every argument becomes a tracer, so it lands in `array_args` and takes the placeholder branch:

```
vjp (eager)               -> placeholders=[]                        warns=0
grad (eager) argnums=0    -> placeholders=[]                        warns=0
jit(grad) argnums=0       -> placeholders=[((3,), 'int32', False)]  warns=1
jit(vjp)  argnums=0       -> placeholders=[((3,), 'int32', False)]  warns=1
```

CPU-only: on the cuda_ipc path the placeholder is a device-side memset, with no NumPy cast.

Not currently reachable from anything in-repo — every array field across `tests/`, `examples/` and `demo/` is `Float32`/`Float64` — but `tesseract-core` exports `Int8`–`Int64`/`UInt8`–`UInt64`, so any gather/indexing Tesseract hits it.

### Steps to reproduce

`gather_tesseract/tesseract_api.py` — differentiable weights, non-differentiable integer indices:

```python
from typing import Any
import numpy as np
from pydantic import BaseModel, Field
[...]

PR #259, opened 2026-09-07
fix: fill discarded derivative slots with each dtype's 0/0 value

jpbrodrick89:
#### Relevant issue or PR

- #258

A derivative call's output tuple has slots that exist only to satisfy its length contract: the gradient of a non-differentiable input, and the tangent of a non-differentiable output. They were filled with `np.full(shape, np.nan, dtype=dtype)`, which is an invalid cast for an integer dtype — NumPy warns and yields `0`. Results are unaffected, but under warnings-as-errors (including this project's own `filterwarnings = ["error"]`) the warning is raised inside the host callback and resurfaces as `JaxRuntimeError: INTERNAL: CpuCallback error calling callback`, which gives a user no way to guess the cause.

#### Description of changes

Non-differentiable output slots are now filled with whatever `0/0` yields for its dtype. That is defined for every dtype we can be handed, so it replaces a per-dtype table with one rule, and keeps poison where poison exists: NaN in every component for the inexact dtypes, zero for the rest. The alternative considered was suppressing the warning while keeping the NaN cast, which is a strictly smaller change but leaves the fill value arbitrary — and leaves the two transports disagreeing on complex.

Two values change, both in slots JAX never propagates (a slot is filled only when its (co)tangent was a symbolic zero, already densified into a constant, so it is never one of the `UndefinedPrimal` positions JAX transposes):

| dtype | before | after |
|---|---|---|
| complex64/128 | `nan+0j` | `nan+nanj` — a consumer reading only `.imag` now sees poison |
| bool | `True` | `False` — NaN was truthy |

Float and integer values are unchanged. Nothing in `tests/`, `examples/` or `demo/` uses complex or bool arrays. The fill is cached per dtype, which makes this faster than the original (see comment below for numbers); the old code cast NaN to the target dtype on every call.

Also widens `unflatten_args`' `array_args` annotation to admit `None`, which the JVP rule already passes per non-differentiated input. Pre-existing but latent: typeguard's default `collection_check_strategy` is `FIRST_ITEM`, so it only surfaces when the untangented input sorts first among the input keys — which `indices` does in the new test Tesseract.

#### Testing done

Full suite: 280 passed (279 before, plus the new test). Confirmed the new test fails on unfixed code with the `RuntimeWarning`.

`tests/gather_tesseract/` is the suite's first integer-array coverage — every array field previously was `Float32`/`Float64`. Its integer input and non-differentiable integer output cover both fill sites in one fixture; the test asserts correct gradients, a `float0` cotangent for the integer input, and an `int32` tangent for the integer output.

Not tested: the GPU path, which fills these slots with a device-side memset rather than NumPy and so cannot hit this warning. That path's values do not follow the rule adopted here — see #239.


jpbrodrick89:
Follow-up on the per-call cost of computing `0/0`: the fill depends only on the dtype, so it's now cached (`functools.cache`), which makes this **faster than the original** rather than slower — the old code was doing an unsafe NaN→int cast on every call.

| shape | dtype | before (`np.full(nan)`) | uncached `0/0` | cached (this PR) |
|---|---|---|---|---|
| `()` | float32 | 0.61 µs | 1.82 µs | **0.51 µs** |
| `()` | int32 | 0.97 µs | 1.37 µs | **0.39 µs** |
| `(1000,)` | float32 | 0.73 µs | 1.96 µs | **0.67 µs** |
| `(1_000_000,)` | float32 | 255 µs | 235 µs | 234 µs |

At large sizes everything is dominated by the memory fill. The cached value is a read-only 0-d array, since it's shared between calls (`np.full` copies from it).

Two notes for reviewers:

- `0-d array / 0-d array` returns a NumPy **scalar**, not an ndarray, so the first cut of this had a wrong return annotation — typeguard caught it across 46 tests. Hence the explicit `np.asarray`.
- On whether output-pruning/DCE would make this moot: it removes the vjp input-side slots (primitive outputs nothing consumes — exactly the DCE target), but **not** the jvp output-side ones, which `jax.jvp` hands back to the caller. Those still need a value whenever the tangent is read, so the cost is worth removing independently of DCE.

Full suite: 280 passed.

dionhaefner:
Couldn't we just fill with zeros and be done with it? JAX isn't allowed to use those values anyway so why overthink it.

jpbrodrick89:
> JAX isn't allowed to use those values anyway

That holds for the **vjp** slot, but not the **jvp** one. A non-differentiable output's tangent is a discarded slot that `jax.jvp` hands straight back to the caller — under `jit`, no GPU, nothing exotic. `tests/pytree_tesseract` already has such an output (`metadata: Array[(2,), Float32]`, non-differentiable), so this needs no new fixture:

```python
def f(alpha_x):
    inputs = {**pytree_tess_inputs, "alpha": {**pytree_tess_inputs["alpha"], "x": alpha_x}}
    return apply_tesseract(pytree_tess, inputs)

@jax.jit
def jvp_jit(x, dx):
    return jax.jvp(f, (x,), (dx,))

primals, tangents = jvp_jit(x, jnp.ones_like(x))
```
Both are keys of the same returned `tangents` pytree — `metadata` is the non-differentiable output, `result` a differentiable one shown for contrast:

```python
tangents["metadata"]   # [nan nan]     non-differentiable -> discarded slot, in the caller's hands
tangents["result"]     # [4. 5. 6.]    differentiable -> a real tangent
```

The reasoning in `_placeholder` ("JAX's transpose machinery never consumes it") is sound for the vjp slot: it is filled only when the cotangent was a symbolic zero, so it is never an `UndefinedPrimal` position and JAX drops it. The jvp slot is not a transpose at all — it is a forward-mode result, and nothing filters it on the way out. One docstring covering two call sites with different visibility is what makes this easy to miss.

## On zeros specifically

Worth separating two things:

1. **This PR does not change the float behaviour.** NaN for inexact dtypes is what `main` already does; #259 only makes the *non-float* dtypes stop disagreeing with it (and stops the int cast warning). So "NaN vs zeros for floats" is a pre-existing decision, not one being introduced here.
2. **Switching to zeros is a one-liner** if we want it — drop the `zero / zero` branch in `_discarded_fill` and always return `zero`. Happy to do that here or as a follow-up.

The trade-off, now that we know the value is observable: with zeros, a user reading the tangent of a non-differentiable output gets `0.0`, which is indistinguishable from "this output genuinely does not depend on the input". That is *correct* when the output really is independent, and a silent lie when the output does depend on the input but the Tesseract declined to differentiate it. NaN can't be mistaken for a real derivative, at the cost of being noisier. I don't feel strongly — but it should be a deliberate choice rather than a side effect of `np.full(..., np.nan, ...)`.

## Tests

`gather_tesseract` now carries a non-differentiable output per dtype class — `count` (int32), `magnitude` (float32), `phase` (complex64) — and `test_discarded_tangents_follow_zero_over_zero` asserts the values end to end, parametrised over `use_jit`:

```
tangent gathered   dtype=float32    [1. 1. 1.]        <- differentiable, unaffected
tangent count      dtype=int32      [0 0 0]           <- no NaN to spell -> zero
tangent magnitude  dtype=float32    [nan nan nan]     <- inexact -> NaN
tangent phase      dtype=complex64  [nan+nanj ...]    <- NaN in *both* components
```

The complex case is the one that fails on `main`: `np.full(shape, np.nan, dtype=complex64)` leaves the imaginary part at a plausible-looking `0`, so a consumer reading only `.imag` sees no poison. (Reverting the fix locally does fail the test, though it trips earlier — on the int32 `RuntimeWarning`.)

Full suite: 282 passed.

Aside: `Complex64`/`Complex128`/`Bool` exist in `tesseract_core.runtime.schema_types` but are not re-exported from `tesseract_core.runtime`, unlike `Float32`/`Int32`. The test Tesseract imports from `schema_types` directly; worth a papercut issue upstream.

tesseract_jax/tesseract_compat.py
```diff
@@ -22,6 +23,31 @@
 # and cannot safely allocate JAX arrays. Use vanilla numpy instead.
 
 
+@functools.cache
+def _discarded_fill(dtype: np.dtype) -> np.ndarray:
```
dionhaefner: Instead of caching this we could just assemble the full dtype registry at import time, and make this a static dict?

tesseract_jax/tesseract_compat.py
```diff
@@ -22,6 +22,66 @@
...
+    dtype = np.dtype(dtype)
+    fill = _DISCARDED_FILL.get(dtype)
+    if fill is None:
+        # A dtype outside the schema set should not reach here, but computing
+        # it beats a KeyError raised inside a host callback, which XLA reports
+        # only as "INTERNAL: CpuCallback error calling callback". Note this
+        # yields zero, not NaN, for an inexact dtype numpy does not recognise
+        # as one (ml_dtypes.bfloat16 among them) -- valid for every dtype,
+        # just weaker poison than the table gives.
+        fill = _compute_discarded_fill(dtype)
```
dionhaefner: This feels overly defensive and my instinct is that we'd rather fail loudly (should not be possible today and I'd like to know tomorrow), but will leave it up to you.

tests/test_api.py
```diff
@@ -755,6 +755,87 @@ def f_raw(d1_):
     np.testing.assert_allclose(jvp(d1, tangent), expected, rtol=1e-5)
 
 
+def test_integer_array_io_is_warning_free(gather_tess):
```
dionhaefner: Is that really the only test? Looks like we're also exercising complex dtypes etc. here, so to me this is filling more than this purpose

tests/test_api.py
```diff
@@ -755,6 +755,87 @@ def f_raw(d1_):
...
+        )
+
+    primals, tangents = jax.jvp(apply_fn, (weights,), (np.ones_like(weights),))
+    np.testing.assert_allclose(primals["gathered"], [1.0, 3.0, 3.0], rtol=1e-6)
+    np.testing.assert_allclose(tangents["gathered"], [1.0, 1.0, 1.0], rtol=1e-6)
+    assert tangents["count"].dtype == jnp.int32
+
+
+@pytest.mark.parametrize("use_jit", [True, False])
+def test_discarded_tangents_follow_zero_over_zero(gather_tess, use_jit):
```
dionhaefner:
```suggestion
def test_discarded_tangent_fill_value(gather_tess, use_jit):
```

tests/test_api.py
```diff
@@ -808,3 +889,30 @@ def test_grad_reaches_a_dotted_dict_key(dict_key_tess, key):
...
+    """The fill table must cover every dtype a Tesseract schema can declare.
+
+    Cross-checked against tesseract-core's own ``dtype`` enum rather than our
+    copy of it, so this fails if tesseract-core grows a dtype and the table is
+    not updated -- at which point the lookup would silently fall back to a
+    computed value instead of the tabulated one.
+    """
+    from tesseract_jax.tesseract_compat import _DISCARDED_FILL
+
+    def _dtype_enums(node):
```
dionhaefner, line 903: One-line docstring would be nice, like "walk an OpenAPI schema and extract all valid dtype enums"

dionhaefner (approved):

tests/test_api.py
jpbrodrick89: no, `test_discarded_tangent_fill_value` is a separate test?

dionhaefner: fair, although it's also asserting results so not really a test of the warning specifically imo

PR #262, opened 2026-09-14
refactor!: dispatch through the primitive in eager mode, always requiring abstract_eval

dionhaefner:
#### Relevant issue or PR

Follow-up to #248, implementing the "option 3" agreed at the end of that thread ([here](https://github.com/pasteurlabs/tesseract-jax/pull/248#issuecomment-5637086730)).

#### Description of changes

Before this change, `apply_tesseract` had two code paths: under a transformation it dispatched through the JAX primitive, and eagerly it called `apply` directly. That made a Tesseract behave differently in the two modes, and let a Tesseract without an `abstract_eval` endpoint run eagerly but not under `jit`.

This sets the primitive's `def_impl` to `dispatch.apply_primitive`, so an eager call traces, compiles and runs the primitive the same way a call under `jit` does. Eager and traced now share one `abstract_eval` + lowering path and behave identically. Two consequences follow:

- **`abstract_eval` is now required in all modes.** A Tesseract without it is rejected eagerly too, with a message pointing at the Tesseract client for running it outside tesseract-jax. To keep both modes consistent we prefer this over silently supporting a degraded eager path (see the discussion in #248 for the alternatives weighed).
- **Errors surface uniformly.** Value-based validation and endpoint exceptions now reach the caller as `JaxRuntimeError` eagerly as well, since the call runs inside the compiled callback either way.

This is a breaking change for anyone relying on the old eager behavior (running a Tesseract without `abstract_eval`, or catching a bare `ValidationError`/`RuntimeError` in eager mode). It should not be released between #248 and this PR.

The stricter dtype check on the compiled callback also surfaced a latent bug in the `nested` test Tesseract: its `abstract_eval` echoed the input dtype and reported `float64` for a field the `OutputSchema` declares `Float32`, disagreeing with what `apply` returns. It now reports the schema dtype.

#### Testing done

Full suite passes locally (320 tests). `ruff check` and `ruff format --check` clean. Tests in `test_nonarray_outputs.py`, `test_exceptions.py` and `test_endtoend.py` updated to the unified eager/traced behavior.


README.md
```diff
@@ -85,12 +85,12 @@ jax.grad(vector_sum)(x, y) # 🎉
...
+- **Additional required endpoints**: Tesseract-JAX requires the [`abstract_eval`](https://docs.pasteurlabs.ai/projects/tesseract-core/latest/content/api/endpoints.html#abstract-eval) Tesseract endpoint to be defined. `apply_tesseract` dispatches through a JAX primitive whether or not a transformation is in play, since an eager call traces, compiles and runs the primitive too, and JAX needs the output shapes in either case. To run a Tesseract that has no `abstract_eval` endpoint, call it directly through the Tesseract client instead. Additionally, many gradient transformations like `jax.grad` require [`vector_jacobian_product`](https://docs.pasteurlabs.ai/projects/tesseract-core/latest/content/api/endpoints.html#vector-jacobian-product) to be defined.
```
jpbrodrick89: I feel this might be unecessary detail. Its tempting to say "to provide output shapes at trace time" but then you have to give the detail about why it traces even outside of jit, a reader scanning the README doesn't need to know this level of detail.

```suggestion
- **Additional required endpoints**: Tesseract-JAX requires the [`abstract_eval`](https://docs.pasteurlabs.ai/projects/tesseract-core/latest/content/api/endpoints.html#abstract-eval) Tesseract endpoint to be defined. (To run a Tesseract that has no `abstract_eval` endpoint, call it directly through the Tesseract client instead.) Additionally, many gradient transformations like `jax.grad` require [`vector_jacobian_product`](https://docs.pasteurlabs.ai/projects/tesseract-core/latest/content/api/endpoints.html#vector-jacobian-product) to be defined.
```

docs/content/get-started.md
```diff
@@ -68,7 +68,7 @@ Now you're ready to jump into our [examples](https://github.com/pasteurlabs/tess
...
+- **Additional required endpoints**: Tesseract-JAX requires the [`abstract_eval`](https://docs.pasteurlabs.ai/projects/tesseract-core/latest/content/api/endpoints.html#abstract-eval) Tesseract endpoint to be defined. `apply_tesseract` dispatches through a JAX primitive whether or not a transformation is in play, since an eager call traces, compiles and runs the primitive too, and JAX needs the output shapes in either case. To run a Tesseract that has no `abstract_eval` endpoint, call it directly through the Tesseract client instead. Additionally, many gradient transformations like `jax.grad` require [`vector_jacobian_product`](https://docs.pasteurlabs.ai/projects/tesseract-core/latest/content/api/endpoints.html#vector-jacobian-product) to be defined.
```
dionhaefner:
```suggestion
- **Additional required endpoints**: Tesseract-JAX requires the [`abstract_eval`](https://docs.pasteurlabs.ai/projects/tesseract-core/latest/content/api/endpoints.html#abstract-eval) Tesseract endpoint to be defined to enable JAX tracing and FFI dispatch. To run a Tesseract that has no `abstract_eval` endpoint, call it directly through the Tesseract client instead. Additionally, many gradient transformations like `jax.grad` require [`vector_jacobian_product`](https://docs.pasteurlabs.ai/projects/tesseract-core/latest/content/api/endpoints.html#vector-jacobian-product) to be defined.
```

docs/content/handling-differentiability.md
```diff
@@ -157,11 +157,12 @@ def f(x):
...
-only return arrays, so a non-array field never enters the computation.
-`apply_tesseract` takes its value from `abstract_eval` and puts it back into the
-output pytree afterwards. Whatever `apply` returns for that field is therefore
-ignored, and `apply_tesseract` warns when the two disagree:
+A JAX primitive can only return arrays, so a non-array field never enters the
+computation. `apply_tesseract` takes its value from `abstract_eval` and puts it
+back into the output pytree afterwards. This holds whether or not a transformation
+is in play, since an eager call traces, compiles and runs the primitive too.
+Whatever `apply` returns for that field is therefore ignored, and `apply_tesseract`
+warns when the two disagree:
```
```suggestion
A JAX primitive can only return arrays, so a non-array field never enters the
computation. `apply_tesseract` takes its value from `abstract_eval` and puts it
back into the output pytree afterwards. Whatever `apply` returns for that field
is therefore ignored, and `apply_tesseract` warns when the two disagree:
```

README.md
```diff
@@ -85,12 +85,12 @@ jax.grad(vector_sum)(x, y) # 🎉
...
 
-- **Non-array outputs come from `abstract_eval` under `jit`**: an `OutputSchema` field that is not an array, such as a `str` or a `bool`, cannot enter a traced computation, so under a JAX transformation its value is the one `abstract_eval` reported and whatever `apply` returns for it is ignored. `apply_tesseract` warns when the two differ; pass `check_static_outputs=False` or set `TESSERACT_JAX_CHECK_STATIC_OUTPUTS=0` to skip the comparison. Without a transformation, `apply` runs directly and its outputs are returned as-is. A field whose value depends on the input values belongs in the schema as an array.
+- **Non-array outputs come from `abstract_eval`**: an `OutputSchema` field that is not an array, such as a `str` or a `bool`, cannot enter a traced computation, so its value is the one `abstract_eval` reported and whatever `apply` returns for it is ignored. `apply_tesseract` warns when the two differ; pass `check_static_outputs=False` or set `TESSERACT_JAX_CHECK_STATIC_OUTPUTS=0` to skip the comparison. A field whose value depends on the input values belongs in the schema as an array.
```
jpbrodrick89, line 93: Fine as is, just attempting to reduce AI smell.

```suggestion
- **Non-array outputs come from `abstract_eval`**: an `OutputSchema` field that is not an array, such as a `str` or a `bool`, cannot enter a traced computation. As such, non-array outputs returned by `apply` are ignored and their value is instead taken from `abstract_eval`. `apply_tesseract` warns when the two differ; pass `check_static_outputs=False` or set `TESSERACT_JAX_CHECK_STATIC_OUTPUTS=0` to skip the comparison. A field whose value depends on the input values belongs in the schema as an array.
```

docs/content/handling-differentiability.md
```diff
@@ -171,16 +172,15 @@ apply, but abstract_eval reported 'reference'. ...
 A field whose value depends on the input values belongs in the schema as an array
 instead.
 
-Without a transformation, `apply` runs directly and its outputs are returned as-is,
-so there is nothing to reconcile and no warning. Under `jit`, a static output cannot
-be returned from the jitted function itself, since JAX has no type for a `str`
-result. Consume it inside the trace as above and return the arrays.
+Under `jit`, a static output cannot be returned from the jitted function itself,
+since JAX has no type for a `str` result. Consume it inside the trace as above and
+return the arrays.
```
```suggestion
```

README.md
dionhaefner:
```suggestion
- **Additional required endpoints**: Tesseract-JAX requires the [`abstract_eval`](https://docs.pasteurlabs.ai/projects/tesseract-core/latest/content/api/endpoints.html#abstract-eval) Tesseract endpoint to be defined to enable JAX tracing and FFI dispatch. To run a Tesseract that has no `abstract_eval` endpoint, call it directly through the Tesseract client instead. Additionally, many gradient transformations like `jax.grad` require [`vector_jacobian_product`](https://docs.pasteurlabs.ai/projects/tesseract-core/latest/content/api/endpoints.html#vector-jacobian-product) to be defined.
```

tests/nested_tesseract/tesseract_api.py
```diff
@@ -97,14 +97,24 @@ def jacobian(inputs: InputSchema, jac_inputs: set[str], jac_outputs: set[str]):
 
 
 def abstract_eval(abstract_inputs):
-    """Calculate output shape of apply from the shape of its inputs."""
+    """Calculate output shape of apply from the shape of its inputs.
+
+    The OutputSchema declares every field as Float32, and apply coerces its
+    outputs to match, so report float32 here regardless of the input dtype.
+    Echoing the input dtype would drift from apply whenever a caller passes a
+    float64 value (e.g. a coerced Python scalar) for a float32 field.
```
jpbrodrick89, line 105: This seems like a bug in tesseract-core, why doesn't abstract_eval coerce its input dtypes to those corresponding to the InputSchema.

tests/test_exceptions.py
```diff
@@ -16,16 +16,12 @@
...
-* under ``jit``, all of them arrive as ``jax.errors.JaxRuntimeError``
-
-Case 1 escapes both, because it fails while tracing rather than in the callback.
-The *message* survives intact either way.
+A Tesseract is dispatched through the JAX primitive whether or not ``jit`` is in
+play -- an eager call traces, compiles and runs it too -- so cases 2-4 run inside
+the compiled callback either way and reach the caller as
+``jax.errors.JaxRuntimeError`` in both modes. Case 1 escapes it, because it fails
+while tracing rather than in the callback. The *message* survives intact
+throughout.
```
jpbrodrick89, line 24: AI smell + unecessary detail

```suggestion
Cases 2-4 run inside the compiled callback and reach the caller as
``jax.errors.JaxRuntimeError``. Case 1 escapes it, because it fails
while tracing rather than in the callback. The *message* survives intact
throughout.
```

tests/test_exceptions.py
```diff
@@ -40,8 +36,12 @@
 
 
 def _expected_error(use_jit):
-    """What a caller sees when the endpoint raises."""
-    return jax.errors.JaxRuntimeError if use_jit else RuntimeError
+    """What a caller sees when the endpoint raises.
+
+    The call runs inside the compiled callback whether or not ``jit`` is in play,
+    so the error arrives as ``JaxRuntimeError`` in both modes.
+    """
```
jpbrodrick89: Unecessary detail

```suggestion
    """What a caller sees when the endpoint raises. """
```

tests/test_exceptions.py
```diff
@@ -75,9 +75,9 @@ def f(v):
 def test_value_based_input_validation_reaches_caller(validating_tess, use_jit):
     """`x > 0` depends on the value, so it can only fail at run time.
 
-    The bound is a schema-level ``AfterValidator``, so it fires while the payload is
-    being parsed rather than in the endpoint body -- which is why this is a
-    ``ValidationError`` and not wrapped in ``RuntimeError`` like cases 3 and 4.
+    The call runs inside the compiled callback whether or not ``jit`` is in play,
+    so the ``ValidationError`` the payload parser raises reaches the caller
+    wrapped in ``JaxRuntimeError`` in both modes.
     """
```
jpbrodrick89: unecessary detail (you might disagree here, my main aim here is to reduce our brain workload when reviewing tests, if there is nothing special to consider here we should not have additional doc string to parse and grok).

```suggestion.
    """`x > 0` depends on the value, so it can only fail at run time."""
```

jpbrodrick89, line 81: unecessary detail (you might disagree here, my main aim here is to reduce our brain workload when reviewing tests, if there is nothing special to consider here we should not have additional doc string to parse and grok).

```suggestion
    """`x > 0` depends on the value, so it can only fail at run time."""
```

tests/test_nonarray_outputs.py
```diff
@@ -3,12 +3,12 @@
...
-taken from `abstract_eval`, carried as static primitive parameters, and put back
-into the output pytree after the bind. These tests check that the arrays are
-untouched either way: the static leaves must not shift, drop or reorder anything
-the gradient path depends on.
+A Tesseract is always dispatched through the JAX primitive, so an eager call
+traces, compiles and runs it just as a call under `jit` would. The primitive can
+only return arrays, so non-array leaves are taken from `abstract_eval`, carried
+as static primitive parameters, and put back into the output pytree after the
+bind. These tests check that the arrays are untouched: the static leaves must not
+shift, drop or reorder anything the gradient path depends on.
```
jpbrodrick89, line 11: Again unecessary detail


```suggestion
A Tesseract is always dispatched through the JAX primitive. The primitive can
only return arrays, so non-array leaves are taken from `abstract_eval`, carried
as static primitive parameters, and put back into the output pytree after the
bind. These tests check that the arrays are untouched: the static leaves must not
shift, drop or reorder anything the gradient path depends on.
```

tesseract_jax/primitive.py
```diff
@@ -989,19 +937,15 @@ def apply_tesseract(
...
         raise ValueError(
-            "Given Tesseract object does not support abstract_eval, "
-            "it is however called in combination with a JAX transformation "
-            "like jit, grad, vmap, or pmap. "
-            "Either remove the transformation or add an abstract_eval endpoint "
-            "to the Tesseract object."
+            "Given Tesseract object does not support abstract_eval, which "
+            "tesseract-jax requires to determine the output shapes of a call. "
+            "Add an abstract_eval endpoint to the Tesseract object, or call it "
+            "directly through the Tesseract client instead of apply_tesseract."
```
jpbrodrick89, line 948:
```suggestion
            "Tesseract object does not support abstract_eval, "
            "tesseract-jax requires it to determine the output shapes of a call. "
            "Add an abstract_eval endpoint to the Tesseract object, or call it "
            "directly through the Tesseract client instead of apply_tesseract."
```

tests/test_exceptions.py
```diff
@@ -40,8 +36,12 @@
...
 
 def _expected_error(use_jit):
-    """What a caller sees when the endpoint raises."""
-    return jax.errors.JaxRuntimeError if use_jit else RuntimeError
+    """What a caller sees when the endpoint raises.
+
+    The call runs inside the compiled callback whether or not ``jit`` is in play,
+    so the error arrives as ``JaxRuntimeError`` in both modes.
+    """
+    return jax.errors.JaxRuntimeError
```
jpbrodrick89, line 44: **Dead code.** This helper now ignores `use_jit` and returns a constant. `test_value_based_input_validation_reaches_caller` already hardcodes `jax.errors.JaxRuntimeError` directly; the two remaining call sites (lines 108 and 124) can do the same and this can go.

```suggestion
```

docs/content/handling-differentiability.md
```diff
@@ -171,16 +170,11 @@ apply, but abstract_eval reported 'reference'. ...
 A field whose value depends on the input values belongs in the schema as an array
 instead.
```
jpbrodrick89: This edit dropped one sentence that is still true and still useful — the `jit`-specific guidance about static outputs not being returnable from the jitted function itself. Only the "without a transformation, `apply` runs directly" sentence needed to go.

```suggestion
Under `jit`, a static output cannot be returned from the jitted function itself,
since JAX has no type for a `str` result. Consume it inside the trace as above and
return the arrays.

A field whose value depends on the input values belongs in the schema as an array
instead.
```

docs/content/get-started.md
```diff
@@ -68,7 +68,7 @@ Now you're ready to jump into our [examples](https://github.com/pasteurlabs/tess
...
+- **Additional required endpoints**: Tesseract-JAX requires the [`abstract_eval`](https://docs.pasteurlabs.ai/projects/tesseract-core/latest/content/api/endpoints.html#abstract-eval) Tesseract endpoint to be defined to enable JAX tracing and FFI dispatch. To run a Tesseract that has no `abstract_eval` endpoint, call it directly through the Tesseract client instead. Additionally, many gradient transformations like `jax.grad` require [`vector_jacobian_product`](https://docs.pasteurlabs.ai/projects/tesseract-core/latest/content/api/endpoints.html#vector-jacobian-product) to be defined.
```
jpbrodrick89: Not this line, but further down this file (line 105, outside the diff so I cannot anchor there): the purity section says

> All of the following require compilation, so they apply under `jax.jit` (including anything jitted internally) and never to eager execution, where each call runs as soon as it is reached:

The conclusion still holds, but the stated reason no longer does — after this PR an eager call compiles too. What actually protects eager callers is that each eager call is its own single-primitive program, not that it is uncompiled. Worth rewording so the mechanism matches the new dispatch model.

tesseract_jax/primitive.py
```diff
@@ -425,27 +389,12 @@ def _raise_if_unimplemented(eval_func: str, client: Jaxeract) -> None:
...
-tesseract_dispatch_p.def_impl(tesseract_dispatch)
+# An eager ``bind`` traces the primitive to a jaxpr, compiles it and runs the
+# compiled program, exactly as a call under ``jit`` would. This routes eager and
+# traced calls through the same ``abstract_eval`` + lowering path, so a Tesseract
+# behaves identically in both. It also means eager use requires an
+# ``abstract_eval`` endpoint, which ``apply_tesseract`` checks up front.
+tesseract_dispatch_p.def_impl(partial(dispatch.apply_primitive, tesseract_dispatch_p))
 
 
 def tesseract_dispatch_lowering(
```
jpbrodrick89: **Dead code** (a few lines below, at line 411): the `if not isinstance(out, tuple): out = (out,)` wrap inside `_dispatch` can never fire now. All four `Jaxeract` endpoints return a tuple unconditionally — `apply` builds `out_data = tuple(...)` since the `output_avals is None` early return went away, and jvp/vjp/jacobian all `return tuple(out)`. A coverage run over the suite confirms line 411 is never executed.

It only existed for the old eager path, where `apply` returned an unflattened pytree. Worth deleting here rather than later: #239 lifts this same block verbatim into `_build_dispatch_closure`, so it would otherwise propagate into the shared CPU/GPU dispatch closure.

docs/content/get-started.md
jpbrodrick89, line 105: This is still true, but I'm not sure if we should reword it (I can't think of a better wording myself).

tesseract_jax/primitive.py
jpbrodrick89, line 411: I think this is dead code now (all endpoints return tuple unconditionally). It only existed for the old eager path, where apply returned an unflattened pytree.

tesseract_jax/primitive.py
```diff
@@ -1012,29 +956,8 @@ def apply_tesseract(
     static_args = tuple(_make_hashable(arg) for arg in static_args)
```
jpbrodrick89, line 956: This currently leads to a discrepancy between eager and jit behaviour (for example it triggers additional compiles every time a concrete array changes value).  I think the correct approach is to treat concrete and tracer arrays on equal footing and also remove `_Hashable` and `make_hashable`. As far as I can tell the only cost we really pay here is:

1. Additional nan allocation in `_discarded_slot` through the transpose rule
2. On accelerators can lead to an additional HDH round trip if concrete arrays are on host.

2 will actually be a positive/wash when #239 lands, its 1 that is a bit more concerning but I think it can be avoided by replacing the `np.full` with a `np.broadcast_to` in `_discarded_slot` (lmk if you want a separate PR for that).

(providing this as pure code because github doesn't seem to display a "suggestion" correctly here)

```
is_static_mask = tuple(not isinstance(arg, jax.Array, np.ndarray) for arg in flat_args)
array_args, static_args = split_args(flat_args, is_static_mask)
```

jpbrodrick89 (approved):
Nice, thanks for getting this in. A few dead code paths missed, we should especially consider whether we should/can drop some of the `use_jit` parameterisation. If we keep it lines 11-13 of test_linearize.py still need editing ("the two settings dispatch through separately registered code paths — tesseract_dispatch_p.def_impl when eager, mlir.register_lowering when staged out")

One way to approach:

- Keep a small deliberate eager set: one smoke test per endpoint family (apply / jvp / vjp / jacobian / vmap), the two abstract_eval-missing rejections, and one error-surfacing case.
- Replace the rest of the use_jit=False variants with a single explicit test that asserts the static classification (assert is_static_mask/operand count, or just that eager and jit agree on values) — that tests the actual remaining difference instead of re-testing it 121 times.
- If you take the is_static_mask fix, eager and jit become literally the same bind, and eager coverage can shrink to a smoke test per endpoint.

Furthermore, the no-abstract_eval ValueError is now raised up front, before any dispatch, yet it's covered by four tests including test_non_abstract_tesseract_apply/_vjp, which spin up a whole tesseract-runtime subprocess (served_non_abstract_tesseract) to assert a pre-dispatch argument check. The two in-process tests in test_nonarray_outputs.py cover it; the fixture and server could go.

tests/nested_tesseract/tesseract_api.py
dionhaefner: This can only happen when using `from_tesseract_api` to pass NumPy arrays, or maybe when passing raw JSON (pre https://github.com/pasteurlabs/tesseract-core/pull/730). Maybe it's unreachable now that 730 is merged, will test.

tesseract_jax/primitive.py
dionhaefner: Went for `np.full`, got a less wasteful fix in the works.

tests/nested_tesseract/tesseract_api.py
dionhaefner: Nope it's a real bug, see https://github.com/pasteurlabs/tesseract-core/issues/759

PR #263, opened 2026-09-15
feat: prune non-differentiable inputs and unused outputs from the VJP

### LINKED ISSUE #4: Fix useless instantiation of zeros
[This line](https://github.com/pasteurlabs/tassadar/blob/c7be35f47c53bda27883ce9658116411b858b175/tassadar/primitive.py#L138) is problematic. We are instantiating zeros and then don't do anything with them. Why, you might ask 🙂  

Let's start with the basics: jax has a funny way of keeping track of which arguments are traced in multi-input functions. Have a look at it yourself: in `test_grad`, when `argnums` is either 0 or 1, if you look at which primals and tangents are sent to, say, the `tesseract_dispatch_jvp_rule` function, you will see that only one argument is non-static (i.e., `len(in_args)==1` and `len(tan_args)==1`), even if the tesseract we are using in the test accepts multiple inputs.

However, when the gradient of a jitted function is calculated, like we do in `test_grad_jit`, even if we try to do the same thing as above we can see that we are passing around two primals and two tangents. Here's what my debugger shows inside `tesseract_dispatch_jvp_rule` for `argnums=2`:
```python3
>>> tan_args
(Zero(ShapedArray(float32[2])), Traced<ShapedArray(float32[2])>with<DynamicJaxprTrace(level=3/0)>)
```
i.e.: we have a tuple of two vectors -- one of which, the one we are not using in differentiation, marked as a symbolic zero.

We could make use of this info when [we build jvp_inputs in the TesseractClient.jacobian_vector_product](https://github.com/pasteurlabs/tassadar/blob/c7be35f47c53bda27883ce9658116411b858b175/tassadar/remote_module.py#L185)... but we don't.
Even if Tesseracts would support calculating jvp only w.r.t. some inputs, this forces the client to always ask for derivatives w.r.t. everything.

dionhaefner:
#### Relevant issue or PR

- Closes #4
- Follow-up to #262
- Related to #209 (which prunes the forward-mode/`jacobian` output request via DCE; this PR handles the reverse-mode/VJP side, so the two are complementary and touch the same output-contract code)

#### Description of changes

Two related changes to how `apply_tesseract` handles a `vector_jacobian_product`, both cutting work the endpoint never needed to do. User-facing gradients are unchanged.

**Omit discarded input slots.** A partial VJP (e.g. `jax.grad`/`jax.vjp` with `argnums`, or `jax.jacrev`) leaves some array inputs non-differentiated. The primitive still returned one output per input, filling a NaN placeholder for each non-differentiable one — a real callback output that JAX allocates and copies into a device buffer, O(n) per input. The VJP arity is now one output per differentiable input; the transpose rule scatters them back into full primal order with `None` (symbolic zero) elsewhere.

**Skip outputs whose cotangent is a symbolic zero (#4).** An output that does not enter the loss carries a symbolic-zero cotangent, but the transpose rule instantiated it to dense zeros and requested a gradient for every non-static output anyway. Besides the wasted work, a NaN in an unused output's gradient (e.g. a normalization undefined at the input) then poisoned the result even though it contributes nothing. A `has_cotangent` mask, captured before instantiation, now prunes those outputs from the request — mirroring how `has_tangent` already prunes inputs.

```python
# Two differentiable outputs, `unsafe` has a NaN gradient at x=0.
def loss(x):
    return apply_tesseract(tess, {"x": x})["safe"].sum()  # `unsafe` unused

jax.grad(loss)(jnp.zeros(3))  # now [2, 2, 2]; previously NaN
```

#### Testing done

Full suite passes (320 tests), including all `grad`/`vjp`/`jacrev`/`jacfwd` and mixed differentiable/non-differentiable input and output cases. New `zero_cotangent_tesseract` fixture with regression tests covering both the pruned and still-requested cases. `ruff check`/`ruff format --check` clean.

Partial VJP on `vectoradd` (differentiate one of two array inputs) is ~25% faster at 10M elements (5.9 ms vs 8.3 ms on `main`), no measurable change at small sizes.


tesseract_jax/primitive.py
```diff
@@ -85,9 +85,16 @@ def tesseract_dispatch_abstract_eval(
     n_primals = params.n_primals
 
     if params.eval_func == "vector_jacobian_product":
-        # We mustn't run forward evaluation of shapes, as out
-        # of vjp has the same shapes as the primals; thus we can return early
-        return tuple(array_args[:n_primals])
+        # A VJP output has the same shape as the primal it differentiates, so we
+        # can read the shapes off the primals without a forward evaluation. Only
+        # differentiable primals (has_tangent) carry a cotangent back; a
+        # non-differentiable input's slot is never consumed by JAX's transpose,
```
jpbrodrick89, line 91: "Differentiable" is misleading here, I don't think we consult the `Differentiable` annotation on this. It is probably  more accurate to use differentiated/tangent-carrying.

tesseract_jax/tesseract_compat.py
```diff
@@ -353,6 +353,15 @@ def vector_jacobian_product(
...
 
+        # Blank out outputs whose cotangent is a symbolic zero (has_cotangent
+        # False) so they drop out of the endpoint request below. It is over the
+        # non-static outputs, so apply it before the static slots are folded in.
+        if params.has_cotangent:
+            cotangents = tuple(
+                c if h else None
+                for c, h in zip(cotangents, params.has_cotangent, strict=True)
+            )
+
```
jpbrodrick89, line 364: Could we also do filtering of `diff_output_path_to_pos` in `_batched_via_jacobian`'s VJP branch?

tesseract_jax/tesseract_compat.py
```diff
@@ -376,37 +385,8 @@ def vector_jacobian_product(
...
-                    )
-                )
-                tan_idx += 1
-
-            # Increment array_idx only for non-static inputs (which appear in array_args)
-            if not params.is_static_mask[all_idx]:
-                array_idx += 1
-
-        return tuple(out)
+        # Only differentiable inputs carry a cotangent back. A non-differentiable
```
jpbrodrick89, line 388:
```suggestion
        # Only differentiated inputs carry a cotangent back. A non-differentiated
```

jpbrodrick89:
```suggestion
        # Only differentiable inputs carry a cotangent back. A non-differentiated
```

tesseract_jax/dispatch_params.py
```diff
@@ -38,6 +38,10 @@ class DispatchParams:
             is carried for it.
         static_output_mask: One flag per output leaf, ``True`` where the leaf is a
             non-array (static) value that never enters the bind.
+        has_cotangent: One flag per non-static output, ``True`` where a non-zero
+            cotangent is carried for it. A ``vector_jacobian_product`` skips the
+            outputs whose cotangent is a symbolic zero, since they add nothing to
+            the input gradients. Empty outside a ``vector_jacobian_product``.
```
jpbrodrick89, line 44:
```suggestion
        has_cotangent: One flag per non-static output, ``True`` where a non-zero
            cotangent is carried for it.
```

tesseract_jax/dispatch_params.py
```diff
@@ -62,6 +66,7 @@ class DispatchParams:
     client: "Jaxeract"
     eval_func: str
     static_output_mask: tuple[bool, ...] = ()
+    has_cotangent: tuple[bool, ...] = ()
```
jpbrodrick89: I'd suggest moving this one up so static_output_mask and static_output_values are still next to each other unless we rely heavily on positional args.

tesseract_jax/dispatch_params.py
jpbrodrick89, line 37: Could consider rewording this to be slightly more accurate:


```suggestion
        has_tangent: One flag per non-static input, True where it carries 
        a nonzero tangent (apply/JVP) or receives a nonzero cotangent (VJP).
```

jpbrodrick89, line 35: wondering if we should rename this `static_input_mask` for symmetry/?

tesseract_jax/primitive.py
```diff
@@ -349,15 +356,29 @@ def tesseract_dispatch_transpose_rule(
...
         *primal_args,
         *cotan_args_,
-        params=params.replace(eval_func="vector_jacobian_product"),
+        params=params.replace(
+            eval_func="vector_jacobian_product", has_cotangent=has_cotangent
+        ),
     )
 
-    return tuple([None] * len(primal_args) + list(vjp))
+    # The bind returns a cotangent only for each differentiable primal
```
jpbrodrick89, line 376:
```suggestion
    # The bind returns a cotangent only for each differentiated primal
```

tesseract_jax/primitive.py
```diff
@@ -631,12 +653,12 @@ def _pad_nans(
         diff_primals,
         jac_cols,
     )
-    diff_pos_set = set(diff_input_path_to_pos.values())
-    grads = _pad_nans(
-        diff_grads,
-        primals,
-        [i in diff_pos_set for i in range(len(primals))],
-    )
+    # A VJP bind returns a gradient only for each differentiable primal; the
```
jpbrodrick89, line 656:
```suggestion
    # A VJP bind returns a gradient only for each differentiated primal; the
```

Again, this should be done earlier (before the dispatch)

jpbrodrick89 (commented):
Looks good, main thing is to make the change before the dispatch in batched via jacobian and improve the comments (including the PR comment).

Follow-ups to consider: DCE (already in the works), removing instantiate zeros itself by modifying bind arity further (hard but might be do-able).

jpbrodrick89:
Also note this PR does not address issue #4, that should be removed from the PR comment to prevent auto-closing

tesseract_jax/primitive.py
```diff
@@ -636,9 +664,16 @@ def _to_batched(arg: Any, ax: Any) -> Any:
 
     # Map each diff output path to its leaf index in the output pytree.
     # ``keys()`` give the path order of the Jacobian's rows; ``values()`` give
-    # the corresponding ``tans`` / ``output_avals`` positions.
+    # the corresponding ``tans`` / ``output_avals`` positions. ``output_flat``
+    # is over non-static outputs in leaf order, so the enumerate index lines up
+    # with ``has_cotangent``. On the VJP path an output whose cotangent is a
+    # symbolic zero adds nothing to the input gradients, so drop it here --
+    # mirroring the transpose rule -- and never request its Jacobian rows.
```
jpbrodrick89: Too waffly, if you want to mention symbolic zeroes feel free to add that back its just not strictly necessary.

```suggestion
    # the corresponding ``tans`` / ``output_avals`` positions dropping unrequested
    # outputs on the reverse-mode path (signalled by `has_cotangent`).
```

tesseract_jax/primitive.py
```diff
@@ -636,9 +664,16 @@ def _to_batched(arg: Any, ax: Any) -> Any:
...
+    is_vjp = params.eval_func == "vector_jacobian_product"
     diff_output_path_to_pos: dict[str, int] = {
-        p: i for i, (p, v) in enumerate(output_flat.items()) if v is not None
+        p: i
+        for i, (p, v) in enumerate(output_flat.items())
+        if v is not None and (not is_vjp or params.has_cotangent[i])
```
jpbrodrick89: Only supported endpoints are `jacobian_vector_product` and `vector_jacobian_product`, so I'd prefer the following for easier parsing:


```suggestion
    is_jvp = params.eval_func == "jacobian_vector_product"
diff_output_path_to_pos: dict[str, int] = {
        p: i
        for i, (p, v) in enumerate(output_flat.items())
        if v is not None and (is_jvp or params.has_cotangent[i])
```

tesseract_jax/primitive.py
```diff
@@ -715,8 +751,11 @@ def _pad_nans(
         return outs, (0,) * len(outs)
 
     # VJP: transpose the outer/inner list structure to iterate
-    # by column (one per diff input).
-    filtered_cots = [tans[pos] for pos in diff_output_path_to_pos.values()]
+    # by column (one per diff input). ``tans`` is already the reduced cotangent
+    # tuple (only ``has_cotangent``-True outputs, symbolic zeros dropped at the
+    # bind), in the same order as the surviving rows of ``diff_output_path_to_pos``,
+    # so take them straight through.
+    filtered_cots = list(tans)
```
jpbrodrick89: I'm not sure the waffle is worth it here, we could just inline `list(tans)` wherever `filtered_cots` is used.

tesseract_jax/primitive.py
```diff
@@ -726,12 +765,12 @@ def _pad_nans(
...
-    diff_pos_set = set(diff_input_path_to_pos.values())
-    grads = _pad_nans(
-        diff_grads,
-        primals,
-        [i in diff_pos_set for i in range(len(primals))],
-    )
+    # A VJP bind returns a gradient only for each differentiated primal; the
+    # abstract_eval declares that shorter arity. Emit the grads in primal
+    # positional order (has_tangent order), omitting the non-diff slots that
+    # JAX's transpose never consumes.
```
jpbrodrick89: Half-tempting to remove the sorting as its not necessary right now.

```suggestion
    # Defensively ensure the grads are in primal positional order.
```

tesseract_jax/tesseract_compat.py
```diff
@@ -496,20 +502,40 @@ def vector_jacobian_product(
...
+        # it before the static slots fold in.
+        if params.has_cotangent:
+            # The bind is set up (in the transpose rule) so its cotangent operands
+            # are exactly the ``has_cotangent``-True slots; a mismatch here would
+            # silently misalign the scatter and corrupt the gradient.
+            assert len(cotangents) == sum(params.has_cotangent)
+            cotan_iter = iter(cotangents)
+            cotangents = tuple(
+                next(cotan_iter) if h else None for h in params.has_cotangent
+            )
```
jpbrodrick89: Too waffly, also consider whether the assert is sufficiently load bearing.

```suggestion
        # Scatter cotangents back to full non-static-output width, inserting
        # ``None`` where the cotangent was a symbolic zero.
        if params.has_cotangent:
            # Guard against misalignment
            assert len(cotangents) == sum(params.has_cotangent)
            cotan_iter = iter(cotangents)
            cotangents = tuple(
                next(cotan_iter) if h else None for h in params.has_cotangent
            )
```

tests/test_jacobian_batching.py
```diff
@@ -378,6 +378,49 @@ def spy(*, inputs, jac_inputs, jac_outputs):
...
+    inputs = {k: jax.tree.map(jnp.asarray, v) for k, v in pytree_tess_inputs.items()}
+
+    captured: dict[str, Any] = {}
+    orig = pytree_tess.jacobian
+
+    def spy(*, inputs, jac_inputs, jac_outputs):
+        captured["jac_outputs"] = sorted(jac_outputs)
+        return orig(inputs=inputs, jac_inputs=jac_inputs, jac_outputs=jac_outputs)
+
+    monkeypatch.setattr(pytree_tess, "jacobian", spy)
```
jpbrodrick89: tempting to spy on `vjp_outputs` again to check consistency with sequential, but agree that might be redundant with`test_output_unused_cotangent_is_not_requested`

jpbrodrick89 (approved):

tests/test_jacobian_batching.py
dionhaefner: I 👓 with my little 👁️

tesseract_jax/primitive.py
dionhaefner: Dropped

PR #265, opened 2026-09-19
doc: document that Tesseracts only support first-order derivatives

### LINKED ISSUE #18: Document restriction to first-order derivatives
Tesseract-JAX cannot be used to compute higher-order derivatives (yet). We should document that.

singhharsh1708:
Adds a "Higher-order derivatives" section to the differentiability page with the two `RuntimeError` messages users actually get, so the page is what turns up when someone searches for them.

Checked against `main`: `jax.hessian`, `jax.grad(jax.grad(f))`, `jax.jacfwd(jax.jacrev(f))` and `jax.jacrev(jax.jacfwd(f))` all raise one of the two, for a Tesseract with a `jacobian` endpoint and for one without (the latter needs a `vmap_method` first). Differentiating the output of `jax.jvp` with respect to the tangent works, hence the last sentence. Docs build clean with `-W`.

Fixes #18.


docs/content/handling-differentiability.md
```diff
@@ -182,3 +182,14 @@ out = apply_tesseract(tess, {"x": x}, check_static_outputs=False)
...
 caller receives are the same either way.
+
+## Higher-order derivatives
+
+Tesseracts only expose first derivatives (`jacobian`, `jacobian_vector_product` and `vector_jacobian_product`), so `apply_tesseract` can be differentiated once with respect to its inputs. Anything that needs a second derivative, such as `jax.hessian(f)`, `jax.grad(jax.grad(f))` or `jax.jacfwd(jax.jacrev(f))`, raises a `RuntimeError`:
+
+```
+RuntimeError: Cannot take higher-order derivatives of 'jacobian'
+RuntimeError: Cannot differentiate a Tesseract derivative endpoint with respect to its primal inputs, as this needs a second derivative.
+```
```
dionhaefner: Would be interesting to show some actual examples that trigger each error. Right now this assumes familiarity with a bunch of concepts (e.g. what a primal is). Some code snippets should make this unambiguous for users.

singhharsh1708: Done, each error now sits under the call that raises it, with a runnable `f`, and "primal" is spelled out.

docs/content/handling-differentiability.md
```diff
@@ -182,3 +182,30 @@ out = apply_tesseract(tess, {"x": x}, check_static_outputs=False)
...
+# RuntimeError: Cannot take higher-order derivatives of 'jacobian'
+
+jax.grad(jax.grad(f))(x)
+# RuntimeError: Cannot differentiate a Tesseract derivative endpoint with respect
+# to its primal inputs, as this needs a second derivative.
+```
+
+Which of the two errors you see depends on how the second derivative is taken. "Primal inputs" means the values `f` is evaluated at, here `x`.
+
+Differentiating with respect to the direction of a derivative instead of `x` works, because a derivative is linear in its direction:
```
```suggestion
Differentiating with respect to a tangent / cotangent works, because a derivative is linear in its direction:
```

dionhaefner (approved):
Thanks @singhharsh1708

